PR #34, opened 2022-12-08
Spherical mesh geometry

mayukh33:


include/flyft/mesh.h
```diff
@@ -1,50 +1,56 @@
 #ifndef FLYFT_MESH_H_
 #define FLYFT_MESH_H_
 
+
```
mphoward, line 4: Try not to add extra white space if you can avoid it. We will use `clang-format` and `pre-commit` eventually to get rid of this, though, so I'm not going to worry too much about it now.

src/cartesian_mesh.cc
```diff
@@ -0,0 +1,8 @@
+#include "flyft/cartesian_mesh.h"
+
+#include <cmath>
```
mphoward, line 3: No need to use this header here
```suggestion
```

src/cartesian_mesh.cc
```diff
@@ -0,0 +1,8 @@
...
+
+double CartesianMesh::volume(int i) const
+    {
+    return step_;
```
mphoward, line 7: Perfect! I think we should start adding the cross-sectional area of the `CartesianMesh` in so that this is a true volume. However, I don't want to make too many changes at once so for now...

Can you add a private member variable to `CartesianMesh` called like `area_` or `cross_area_` (whatever you prefer), give it a default value of `1.0` in the constructor, then use it here like:
```suggestion
    return area_*step_;
```

mphoward (changes requested):
Good start! Here are some suggestions. When you are ready, ping me on Slack for ideas for next steps to take.

src/cartesian_mesh.cc
```diff
@@ -0,0 +1,26 @@
+#include "flyft/cartesian_mesh.h"
+
+double CartesianMesh::CartesianMesh(double L, int shape)
+    : Mesh(L,shape), area_(1.0)
```
mphoward, line 4:
```suggestion
    : CartesianMesh(L,shape,0)
```

src/cartesian_mesh.cc
```diff
@@ -0,0 +1,26 @@
...
+    {
+    }
+
+double CartesianMesh::CartesianMesh(double L,int shape, double origin)
+    : Mesh(L,shape,origin)
```
mphoward, line 9:
```suggestion
    : Mesh(L,shape,origin), area_(1.0)
```

src/cartesian_mesh.cc
```diff
@@ -0,0 +1,26 @@
...
+    {
+    }
+
+double CartesianMesh::CartesianMesh(int shape,double step)
+    : Mesh(shape,step,0)
```
mphoward, line 14:
```suggestion
    : CartesianMesh(shape,step,0)
```

src/cartesian_mesh.cc
```diff
@@ -0,0 +1,26 @@
...
+    {
+    }
+
+double CartesianMesh::CartesianMesh(int shape, double step, double origin)
+    : Mesh(shape,step,origin)
```
mphoward, line 19:
```suggestion
    : Mesh(shape,step,origin), area_(1.0)
```

src/boublik_hard_sphere_functional.cc
```diff
@@ -114,7 +114,7 @@ static void computeFunctionalPure(int idx,
         // compute free energy
         if (compute_value)
             {
-            energy = mesh.step()*rho*eta*(4.-3.*eta)/vf_2;
+            energy = mesh->step()*rho*eta*(4.-3.*eta)/vf_2;
```
mphoward, line 117: Should this be volume?

src/crank_nicolson_integrator.cc
```diff
@@ -76,9 +76,9 @@ void CrankNicolsonIntegrator::step(std::shared_ptr<Flux> flux,
             #pragma omp parallel for schedule(static) default(none) firstprivate(timestep,mesh,alpha,tol) \
                 shared(next_rho,next_j,last_rho,last_rate,converged)
             #endif
-            for (int idx=0; idx < mesh.shape(); ++idx)
+            for (int idx=0; idx < mesh->shape(); ++idx)
                 {
-                const double next_rate = (next_j(idx)-next_j(idx+1))/mesh.step();
+                const double next_rate = (next_j(idx)-next_j(idx+1))/mesh->step();
```
mphoward, line 81: This looks like it needs to be modified similarly as above, no?

Should we consider adding a method to the `Mesh` that will do the surface integral over a site to reduce code duplication? Something like this might be nice to give two different ways to write this:
```
double integrateSurface(int idx, double j_lo, double j_hi) const
    {
    return area(idx)*j_lo - area(idx+1)*j_hi;
    }
double integrateSurface(int idx, const DataView<double>& j) const
    {
    return integrateSurface(idx, j(idx), j(idx+1));
    }
```
You could then use it like:
```
const double next_rate = mesh->integrateSurface(idx, next_j)/mesh->volume(idx);
```

We could add a similar `integrateVolume` too, if we wanted to:
```
double integrateVolume(int idx, double f) const
    {
    return volume(idx)*f;
    }
double integrateVolume(int idx, const DataView<double>& f) const
    {
    return integrateVolume(idx, f(idx));
    }
```

This would match nicely when you went to implement the `gradient` method, which would be virtual.

src/explicit_euler_integrator.cc
```diff
@@ -25,9 +25,9 @@ void ExplicitEulerIntegrator::step(std::shared_ptr<Flux> flux,
         #ifdef FLYFT_OPENMP
         #pragma omp parallel for schedule(static) default(none) firstprivate(timestep,mesh) shared(rho,j)
         #endif
-        for (int idx=0; idx < mesh.shape(); ++idx)
+        for (int idx=0; idx < mesh->shape(); ++idx)
             {
-            const auto rate = (j(idx)-j(idx+1))/mesh.step();
+            const auto rate = (mesh->area(idx)/mesh->volume(idx))*j(idx)-(mesh->area(idx+1)/mesh->volume(idx))*j(idx+1);
```
mphoward, line 30: Revise as suggested above.

src/external_potential.cc
```diff
@@ -65,7 +66,7 @@ void ExternalPotential::_compute(std::shared_ptr<State> state, bool compute_valu
                     }
                 else if (compute_value)
                     {
-                    value_ += mesh.step()*rho*V;
+                    value_ += mesh->volume(idx)*rho*V;
```
mphoward, line 69: With the integration method, this would look like:
```
value_ += mesh->integrateVolume(idx, rho*V);
```

src/grand_potential.cc
```diff
@@ -71,12 +71,12 @@ void GrandPotential::_compute(std::shared_ptr<State> state, bool compute_value)
...
             shared(rho,d,compute_value) reduction(-:constraint_value)
             #endif
-            for (int idx=0; idx < mesh.shape(); ++idx)
+            for (int idx=0; idx < mesh->shape(); ++idx)
                 {
                 d(idx) -= mu_bulk;
                 if (compute_value)
                     {
-                    constraint_value -= mesh.step()*mu_bulk*rho(idx);
+                    constraint_value -= mesh->volume(idx)*mu_bulk*rho(idx);
```
mphoward, line 79: But this would look like:
```suggestion
                    constraint_value -= mu_bulk*mesh->integrateVolume(idx, rho);
```

mphoward (changes requested):
Good work! I had some suggestions for how we might reduce a little code duplication (and effort to make changes in future) by making the volume / surface integrals over a site be methods of the mesh. Could you try implementing these next?

include/flyft/mesh.h
```diff
@@ -1,6 +1,9 @@
 #ifndef FLYFT_MESH_H_
 #define FLYFT_MESH_H_
 
+#include <memory>
+#include "data_view.h"
```
mphoward, line 5: You need to include flyft headers with the `flyft/` prefix:
```suggestion
#include "flyft/data_view.h"

#include <memory>
```
Also, headers should come in the following order:
1. This project
2. External libraries
3. Standard library
Separate each group with a line break.

include/flyft/mesh.h
```diff
@@ -34,17 +37,28 @@ class Mesh
...
 
         double asLength(int shape) const;
+        
+        double integrateSurface(int idx, double j_lo, double j_hi) const;
+        
+        double integrateSurface(int idx, const DataView<double>& j) const;
+        
+        double integrateVolume(int idx, double f) const;
 
+        double integrateVolume(int idx, const DataView<double>& f) const;
```
mphoward, line 47: For readability, group overloaded functions together
```suggestion
        double integrateSurface(int idx, double j_lo, double j_hi) const;
        double integrateSurface(int idx, const DataView<double>& j) const;
        
        double integrateVolume(int idx, double f) const;
        double integrateVolume(int idx, const DataView<double>& f) const;
```

src/boublik_hard_sphere_functional.cc
```diff
@@ -69,7 +69,7 @@ static void computeFunctionalMixture(int idx,
         // compute free energy
         if (compute_value)
             {
-            energy = mesh.step()*(6./M_PI)*((xi2_3/xi3_2-xi[0])*logvf + 3.*xi[1]*xi[2]/vf + xi2_3/(xi[3]*vf_2));
+            energy = mesh->volume(idx)*(6./M_PI)*((xi2_3/xi3_2-xi[0])*logvf + 3.*xi[1]*xi[2]/vf + xi2_3/(xi[3]*vf_2));
```
mphoward, line 72:
```suggestion
            energy = mesh->integrateVolume(idx, (6./M_PI)*((xi2_3/xi3_2-xi[0])*logvf + 3.*xi[1]*xi[2]/vf + xi2_3/(xi[3]*vf_2));
```

src/boublik_hard_sphere_functional.cc
```diff
@@ -114,7 +114,7 @@ static void computeFunctionalPure(int idx,
         // compute free energy
         if (compute_value)
             {
-            energy = mesh.step()*rho*eta*(4.-3.*eta)/vf_2;
+            energy = mesh->volume(idx)*rho*eta*(4.-3.*eta)/vf_2;
```
mphoward, line 117:
```suggestion
            energy = mesh->integrateVolume(idx, rho*eta*(4.-3.*eta)/vf_2);
```

src/crank_nicolson_integrator.cc
mphoward, line 81: Shouldn't this be:
```suggestion
                const double next_rate = mesh->integrateSurface(idx, next_j)/mesh->volume(idx);
```

src/explicit_euler_integrator.cc
```diff
@@ -25,9 +25,9 @@ void ExplicitEulerIntegrator::step(std::shared_ptr<Flux> flux,
         #ifdef FLYFT_OPENMP
         #pragma omp parallel for schedule(static) default(none) firstprivate(timestep,mesh) shared(rho,j)
         #endif
-        for (int idx=0; idx < mesh.shape(); ++idx)
+        for (int idx=0; idx < mesh->shape(); ++idx)
             {
-            const auto rate = (j(idx)-j(idx+1))/mesh.step();
+            const auto rate = mesh->integrateSurface(idx,j(idx),j(idx+1))/mesh->volume(idx);
```
mphoward, line 30: Since `j` is a DataView, you should be able to use that overload
```suggestion
            const auto rate = mesh->integrateSurface(idx, j))/mesh->volume(idx);
```

src/grand_potential.cc
```diff
@@ -71,12 +71,12 @@ void GrandPotential::_compute(std::shared_ptr<State> state, bool compute_value)
...
             shared(rho,d,compute_value) reduction(-:constraint_value)
             #endif
-            for (int idx=0; idx < mesh.shape(); ++idx)
+            for (int idx=0; idx < mesh->shape(); ++idx)
                 {
                 d(idx) -= mu_bulk;
                 if (compute_value)
                     {
-                    constraint_value -= mesh.step()*mu_bulk*rho(idx);
+                    constraint_value -= mu_bulk*mesh->integrateVolume(idx,rho(idx));
```
mphoward, line 79: Same here, `rho` is a `DataView` so that overload should work:
```suggestion
                    constraint_value -= mu_bulk*mesh->integrateVolume(idx, rho);
```

src/ideal_gas_functional.cc
```diff
@@ -23,7 +23,7 @@ static void computeFunctional(int idx,
         d(idx) = std::log(vol*rho);
         if (compute_value)
             {
-            energy = mesh.step()*rho*(d(idx)-1.);
+            energy = mesh->volume(idx)*rho*(d(idx)-1.);
```
mphoward, line 26:
```suggestion
            energy = mesh->integrateVolume(idx, rho*(d(idx)-1.));
```

src/implicit_euler_integrator.cc
```diff
@@ -61,9 +61,9 @@ void ImplicitEulerIntegrator::step(std::shared_ptr<Flux> flux,
             #ifdef FLYFT_OPENMP
             #pragma omp parallel for schedule(static) default(none) firstprivate(timestep,mesh,alpha,tol) shared(next_rho,next_j,last_rho,converged)
             #endif
-            for (int idx=0; idx < mesh.shape(); ++idx)
+            for (int idx=0; idx < mesh->shape(); ++idx)
                 {
-                const double next_rate = (next_j(idx)-next_j(idx+1))/mesh.step();
+                const double next_rate = mesh->integrateSurface(idx,next_j(idx),next_j(idx+1))/mesh->volume(idx);
```
mphoward, line 66:
```suggestion
                const double next_rate = mesh->integrateSurface(idx, next_j)/mesh->volume(idx);
```

src/picard_iteration.cc
```diff
@@ -63,9 +63,8 @@ bool PicardIteration::solve(std::shared_ptr<GrandPotential> grand, std::shared_p
                         eff_energy += V(idx);
                         }
                     rho_tmp(idx) = std::exp(-eff_energy);
-                    sum += rho_tmp(idx);
+                    sum += mesh->integrateVolume(idx, rho(idx));
```
mphoward, line 66:
```suggestion
                    sum += mesh->integrateVolume(idx, rho);
```

src/rosenfeld_fmt.cc
```diff
@@ -263,9 +263,9 @@ void RosenfeldFMT::_compute(std::shared_ptr<State> state, bool compute_value)
             #ifdef FLYFT_OPENMP
             #pragma omp parallel for schedule(static) default(none) firstprivate(mesh) shared(phi) reduction(+:value_)
             #endif
-            for (int idx=0; idx < mesh.shape(); ++idx)
+            for (int idx=0; idx < mesh->shape(); ++idx)
                 {
-                value_ += mesh.step()*phi(idx);
+                value_ += mesh->integrateVolume(idx, phi(idx));
```
mphoward, line 268:
```suggestion
                value_ += mesh->integrateVolume(idx, phi);
```

src/virial_expansion.cc
```diff
@@ -32,14 +32,14 @@ static void computeFunctional(int idx,
         }
     if (compute_value)
         {
-        value += mesh.step()*(factor*Bij*rhoi*rhoj);
+        value += mesh->volume(idx)*(factor*Bij*rhoi*rhoj);
```
mphoward, line 35:
```suggestion
        value += mesh->integrateVolume(idx, factor*Bij*rhoi*rhoj);
```

mphoward (changes requested):
Looks like you implemented the integrate methods correctly, but you missed quite a few spots applying them in the code. I think I got them all, so please apply these changes and then double check that there aren't any spots I missed!

include/flyft/mesh.h
```diff
@@ -46,6 +50,10 @@ class Mesh
         double integrateVolume(int idx, const DataView<double>& f) const;
         double integrateVolume(int idx, double f) const;
         
+        double linearInterpolate(int idx, const DataView<const double>& rho) const;
+        double linearInterpolate(int idx, const DataView<double>& rho) const;
+        double linearInterpolate(int idx, double rho_lo, double rho_hi) const;
```
mphoward, line 55: I would rather call these `f` than `rho`, as I think it is more readable. Also, I would just call these `interpolate` so that it gives deriving classes the freedom to override the style of interpolation with whatever it thinks is best. For example:
```
double interpolate(int idx, double f_lo, double f_hi) const;
```

src/brownian_diffusive_flux.cc
```diff
@@ -37,17 +37,17 @@ static double calculateFlux(int idx,
     if (!no_flux)
         {
         // average density at left edge
-        auto rho_avg = 0.5*(rho(idx-1)+rho(idx));
+        auto rho_avg = mesh->linearInterpolate(idx,rho);
```
mphoward, line 40: This would look like:
```suggestion
        auto rho_avg = mesh->interpolate(idx,rho);
```

src/mesh.cc
```diff
@@ -35,6 +35,16 @@ int Mesh::bin(double x) const
     return static_cast<int>((x-origin_)/step_);
     }
 
+double Mesh::lo_edge(int i) const
+    {
+    return origin_+static_cast<double>(i)*step_;
```
mphoward, line 40: Perfect!

src/spherical_mesh.cc
```diff
@@ -26,11 +26,11 @@ double SphericalMesh::volume(int i) const
 
 double SphericalMesh::gradient(int idx, double f_lo, double f_hi) const
     {
-    if(coordinate(idx) == 0){
-        return 3*(f_hi-f_lo)/step_;
+    if(idx == 0){
```
mphoward, line 29: I think you need an extra check here that `origin_ == 0.`, as this condition will be met on multiple domains:
```
if (idx == 0 && origin_ == 0.)
```
Alternatively, you can use the lower bound method that you added, at the expense of an extra multiplication operation. Hence, I would probably favor this version.

mphoward (changes requested):
Looks good! A couple small things I noticed.

python/flyft/_flyft/spherical_mesh.cc
```diff
@@ -0,0 +1,12 @@
+#include "_flyft.h"
+#include "flyft/spherical_mesh.h"
+
+void bindSphericalMesh(py::module_& m)
+    {
+    using namespace flyft;
+
+    py::class_<SphericalMesh,std::shared_ptr<SphericalMesh>,Mesh>(m, "SphericalMesh")
+        .def(py::init<int,double,double>())
```
mphoward, line 9: Is this the only constructor of the `SphericalMesh`, or are there others?

python/flyft/_flyft/state.cc
```diff
@@ -8,10 +8,9 @@ void bindState(py::module_& m)
     using namespace flyft;
 
     py::class_<State,std::shared_ptr<State>>(m, "State")
-        .def(py::init<double,int,const std::string&>())
-        .def(py::init<double,int,const std::vector<std::string>&>())
-        .def(py::init<double,int,const std::string&,std::shared_ptr<Communicator>>())
-        .def(py::init<double,int,const std::vector<std::string>&,std::shared_ptr<Communicator>>())
+        .def(py::init<std::shared_ptr<ParallelMesh>,const std::string&>())
+        .def(py::init<std::shared_ptr<ParallelMesh>, std::vector<std::string>&>())
```
mphoward, line 12: I think this should be const?
```suggestion
        .def(py::init<std::shared_ptr<ParallelMesh>, const std::vector<std::string>&>())
```

python/flyft/_flyft/state.cc
```diff
@@ -8,10 +8,9 @@ void bindState(py::module_& m)
...
+        .def(py::init<const State&>())
```
mphoward, line 13: Why was this added? We don't want to expose a copy constructor.
```suggestion
```

mphoward (changes requested):
Looks good! Here are a couple small things. Did the code compile?

python/flyft/_flyft/spherical_mesh.cc
mayukh33: The spherical mesh did not have the other constructors. I added them now. The code is compiling!

include/flyft/spherical_mesh.h
```diff
@@ -8,11 +8,16 @@ namespace flyft
 class SphericalMesh : public Mesh
     {
     public:
-        SphericalMesh(int shape, double step,double origin);
+        SphericalMesh(double L, int shape);
+        SphericalMesh(double L, int shape, double origin);
```
mphoward, line 12: Use R rather than L, since this is a sphere?
```suggestion
        SphericalMesh(double R, int shape);
        SphericalMesh(double R, int shape, double origin);  
```

src/spherical_mesh.cc
```diff
@@ -4,6 +4,25 @@
 
 namespace flyft
 {
+SphericalMesh::SphericalMesh(double L, int shape)
```
mphoward, line 7: Apply the rename to R here.

python/flyft/_flyft/_flyft.cc
```diff
@@ -10,6 +10,8 @@ void bindDoubleParameters(py::module_&);
 void bindCommunicator(py::module_&);
 void bindField(py::module_&);
 void bindMesh(py::module_&);
+void bindSphericalMesh(py::module_&);
+void bindCartesianMesh(py::module_&);
```
mphoward, line 14: You need to make sure to add `cartesian_mesh.cc` and `spherical_mesh.cc` to `python/flyft/CMakeLists.txt` or they will not be compiled.

include/flyft/mesh.h
```diff
@@ -34,17 +37,41 @@ class Mesh
         int asShape(double dx) const;
 
         double asLength(int shape) const;
+        
+        double lower_bound(int i) const;
```
mphoward, line 41: Now that these are called `lower_bound` and `upper_bound` (which I like alot!), would it make sense to rename `coordinate` to `center`? Then it would be clearer what this quantity represents.

mphoward (changes requested):
Looks great! A couple small comments to resolve now, then we will move onto the pure Python code!

include/flyft/mesh.h
mayukh33: That makes sense. I will make that change.

python/flyft/state.py
```diff
@@ -51,36 +51,50 @@ def __init__(self, L, shape, origin=0.):
...
 
     @property
-    def coordinates(self):
-        if not hasattr(self, '_coordinates'):
-            self._coordinates = np.array([self._self.coordinate(i) for i in range(self.shape)])
-        return self._coordinates
+    def center(self):
+        if not hasattr(self, '_center'):
+            self._center = np.array([self._self.center(i) for i in range(self.shape)])
+        return self._center
```
mphoward, line 57: We currently pluralize anything that's an array, so this should be:
```suggestion
    def centers(self):
        if not hasattr(self, '_centers'):
            self._centers = np.array([self._self.center(i) for i in range(self.shape)])
        return self._centers
```

python/flyft/state.py
```diff
@@ -51,36 +51,50 @@ def __init__(self, L, shape, origin=0.):
...
     shape = mirror.Property()
     step = mirror.Property()
 
+class CartesianMesh(mirror.Mirror,mirrorclass=_flyft.CartesianMesh):
+    def __init__(self,L,shape,area = 1.):      
+        Mesh.__init__(self,L,shape) 
+
+class SphericalMesh(mirror.Mirror,mirrorclass=_flyft.SphericalMesh):
+    def __init__(self,R,shape):
+        Mesh.__init__(self,R,shape)
```
mphoward, line 70: Do the same here as for `CartesianMesh`

python/flyft/state.py
```diff
@@ -51,36 +51,50 @@ def __init__(self, L, shape, origin=0.):
...
     full = mirror.Property()
     local = mirror.Property()
 
 class State(mirror.Mirror,mirrorclass=_flyft.State):
-    def __init__(self, L, shape, types, communicator=None):
+    def __init__(self, mesh, communicator=None):
         # cast type list into vector
+        super().__init__(mesh, communicator)
+        self._communicator = communicator
+
```
mphoward, line 86: You should only call the constructor once, and it is done correctly below. Remove this.
```suggestion
```

python/flyft/state.py
```diff
@@ -51,36 +51,50 @@ def __init__(self, L, shape, origin=0.):
...
+        communicator = Communicator()
+        super().__init__(mesh, communicator)
         self._communicator = communicator
 
     full = mirror.Property()
     local = mirror.Property()
 
 class State(mirror.Mirror,mirrorclass=_flyft.State):
-    def __init__(self, L, shape, types, communicator=None):
+    def __init__(self, mesh, communicator=None):
```
mphoward, line 82: The `State` constructor needs a `ParallelMesh` and the `types`, but not the `communicator`:
```suggestion
    def __init__(self, mesh, types):
```

python/flyft/state.py
```diff
@@ -51,36 +51,50 @@ def __init__(self, L, shape, origin=0.):
...
         # cast type list into vector
+        super().__init__(mesh, communicator)
+        self._communicator = communicator
+    
         if isinstance(types, str):
             types = (types,)
+        
+        communicator = Communicator()
```
mphoward, line 90: Don't need this either, not sure why it is here.
```suggestion
```

python/flyft/state.py
```diff
@@ -51,36 +51,50 @@ def __init__(self, L, shape, origin=0.):
...
 
         # initialize using communicator, if specified
         if communicator is None:
-            super().__init__(L,shape,_flyft.VectorString(types))
+            super().__init__(mesh,_flyft.VectorString(types))
```
mphoward, line 94: You are only going to need this constructor now. Remove the `if-else`, and only keep this branch

python/flyft/state.py
```diff
@@ -51,10 +51,10 @@ def __init__(self, L, shape, origin=0.):
         super().__init__(L, shape, origin)
 
     @property
-    def center(self):
-        if not hasattr(self, '_center'):
-            self._center = np.array([self._self.center(i) for i in range(self.shape)])
-        return self._center
+    def centers(self):
+        if not hasattr(self, '_centers'):
+            self._centers = np.array([self._self.centers(i) for i in range(self.shape)])
```
mphoward, line 56: This method is not plural, it will crash if you actually run code with it:
```suggestion
            self._centers = np.array([self._self.center(i) for i in range(self.shape)])
```

python/flyft/state.py
```diff
@@ -63,11 +63,11 @@ def center(self):
 
 class CartesianMesh(mirror.Mirror,mirrorclass=_flyft.CartesianMesh):
     def __init__(self,L,shape,area = 1.):      
-        Mesh.__init__(self,L,shape) 
+        super().__init__(L,shape,area)
```
mphoward, line 66: You still need to delete the `__init__` method for the `Mesh` above, or I suspect this will not work.

python/flyft/state.py
```diff
@@ -79,23 +79,12 @@ def __init__(self,mesh, communicator):
     local = mirror.Property()
 
 class State(mirror.Mirror,mirrorclass=_flyft.State):
-    def __init__(self, mesh, communicator=None):
-        # cast type list into vector
-        super().__init__(mesh, communicator)
-        self._communicator = communicator
-    
+    def __init__(self, mesh, types):
+        super().__init__(mesh,_flyft.VectorString(types))
```
mphoward, line 83: This was in the right spot below because it needs to come after `types` is cast to a tuple. Please put it back down there, you only need to remove the `if-else` logic and keep this line.

mphoward (changes requested):

mphoward (commented):
These changes all look good, onto the unit tests!

python/flyft/state.py
```diff
@@ -47,44 +47,45 @@ def __getitem__(self, key):
...
+class SphericalMesh(mirror.Mirror,mirrorclass=_flyft.SphericalMesh):
+    def __init__(self,R,shape):
+        super().__init__(R,shape)
+
 class ParallelMesh(mirror.Mirror,mirrorclass=_flyft.ParallelMesh):
-    def __init__(self, L, shape, communicator):
-        super().__init__(Mesh(L,shape), communicator)
+    def __init__(self,mesh):
+        communicator = Communicator()
+        super().__init__(mesh)
```
mphoward, line 73: You shouldn't have removed the communicator here, that might be causing your error:
```suggestion
        super().__init__(mesh, communicator)
```

src/cartesian_mesh.cc
```diff
@@ -0,0 +1,45 @@
...
+    {
+    }
+
+CartesianMesh::CartesianMesh(double L,int shape, double area, double origin)
+    : Mesh(L,shape,origin), area_(area)
+    {
+    }
+
+CartesianMesh::CartesianMesh(int shape, double step, double area)
+    : CartesianMesh(shape,step,0,area)
```
mphoward, line 17:
```suggestion
    : CartesianMesh(shape,step,area,0)
```

src/cartesian_mesh.cc
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+CartesianMesh::CartesianMesh(int shape, double step,double area, double origin)
+    : Mesh(shape,step,origin),area_(area)
+    {
+    }
+
+std::shared_ptr<Mesh> CartesianMesh::slice(int start, int end) const
+    {
+    return std::shared_ptr<Mesh>(new CartesianMesh(end-start,step_,lower_bound(start)));
```
mphoward, line 28:
```suggestion
    return std::shared_ptr<Mesh>(new CartesianMesh(end-start,step_,area_,lower_bound(start)));
```

src/picard_iteration.cc
```diff
@@ -63,7 +63,7 @@ bool PicardIteration::solve(std::shared_ptr<GrandPotential> grand, std::shared_p
                         eff_energy += V(idx);
                         }
                     rho_tmp(idx) = std::exp(-eff_energy);
-                    sum += mesh->integrateVolume(idx, rho);
+                    sum += mesh->integrateVolume(idx, rho_tmp);
```
mphoward, line 66: Good find! Was a test failing because of this?

python/tests/test_explicit_euler_integrator.py
```diff
@@ -73,7 +73,7 @@ def test_advance(state,grand,ig,linear,bd,euler):
 
 @pytest.mark.parametrize("adapt",[False,True])
 def test_sine(adapt,euler):
-    mesh = flyft.state.ParallelMesh(flyft.state.CartesianMesh(10.0,20,1))
+    mesh = flyft.state.ParallelMesh(flyft.state.CartesianMesh(2.,100,1))
```
mphoward, line 76: Why this change?

mphoward (changes requested):

src/picard_iteration.cc
mayukh33: Yes! The picard iteration test was failing

include/flyft/state.h
```diff
@@ -19,10 +19,8 @@ class State : public TrackedObject
     {
     public:
         State() = delete;
-        State(double L, int shape, const std::string& type);
-        State(double L, int shape, const std::vector<std::string>& types);
-        State(double L, int shape, const std::string& type, std::shared_ptr<Communicator> comm);
-        State(double L, int shape, const std::vector<std::string>& types, std::shared_ptr<Communicator> comm);
+        State(std::shared_ptr<ParallelMesh> mesh, const std::string& type);
+        State(std::shared_ptr<ParallelMesh> mesh, const std::vector<std::string>& types);
```
mphoward, line 23: Would it make sense to go to an intermediate solution like this?
```c++
State(std::shared_ptr<Mesh> mesh, const std::string& type);
State(std::shared_ptr<Mesh> mesh, const std::vector<std::string>& type);
State(std::shared_ptr<Mesh> mesh, const std::string& type, std::shared_ptr<Communicator> comm);
State(std::shared_ptr<Mesh> mesh, const std::vector<std::string>& type, std::shared_ptr<Communicator> comm);
```

python/flyft/_flyft/mesh.cc
```diff
@@ -15,9 +15,14 @@ class MeshTrampoline : public Mesh
         PYBIND11_OVERRIDE_PURE(double, Mesh, area, i);
         }
         
+        double volume() const override 
+        {
+        PYBIND11_OVERLOAD_PURE_NAME(double, Mesh, "total_volume", volume, );
```
mphoward, line 20: How about we just call this `volume` (and remove the extra `,`:
```suggestion
        PYBIND11_OVERLOAD_PURE(double, Mesh, volume);
```

python/flyft/_flyft/mesh.cc
```diff
@@ -50,5 +55,7 @@ void bindMesh(py::module_& m)
         .def("bin", &Mesh::bin)
         .def("lower_bound",&Mesh::lower_bound)
         .def("upper_bound",&Mesh::upper_bound)
+        .def("total_volume",py::overload_cast<>(&Mesh::volume, py::const_))
```
mphoward, line 58: We could decide to make this one a property and read only, it would make sense. I think this would work:
```suggestion
        .def_property_readonly("volume", py::overload_cast<>(&Mesh::volume, py::const_))
```

python/flyft/state.py
```diff
@@ -53,12 +53,23 @@ def centers(self):
         if not hasattr(self, '_centers'):
             self._centers = np.array([self._self.center(i) for i in range(self.shape)])
         return self._centers
+    
+        
+    def volume(self, bin = None):
+        if bin is None:
+            v = self._self.total_volume()
```
mphoward, line 60: If changes above work, this would be:
```suggestion
            v = self._self.volume
```

python/flyft/state.py
```diff
@@ -53,12 +53,23 @@ def centers(self):
         if not hasattr(self, '_centers'):
             self._centers = np.array([self._self.center(i) for i in range(self.shape)])
         return self._centers
+    
+        
+    def volume(self, bin = None):
```
mphoward, line 58:
```suggestion
    def volume(self, bin=None):
```

python/tests/conftest.py
```diff
@@ -18,14 +19,31 @@ def grand():
...
+def spherical_mesh():
+    return flyft.state.SphericalMesh(10.0,20)
+
+@pytest.fixture(params=[lazy_fixture("cartesian_mesh"), lazy_fixture("spherical_mesh")])
+def mesh(request):
+    return request.param
+
+@pytest.fixture  
+def volumes(mesh):
+    return mesh.volume()
```
mphoward, line 35: This should not be a fixture
```suggestion
```

python/tests/test_boublik_hard_sphere.py
```diff
@@ -37,9 +39,8 @@ def test_diameters(bmcsl):
     assert bmcsl._self.diameters['A'] == pytest.approx(1.5)
     assert bmcsl._self.diameters['B'] == pytest.approx(2.5)
 
-def test_compute(bmcsl,binary_state):
+def test_compute(bmcsl,binary_state,volumes):
```
mphoward, line 42:
```suggestion
def test_compute(bmcsl,binary_state):
```

python/tests/test_boublik_hard_sphere.py
```diff
@@ -50,7 +51,7 @@ def test_compute(bmcsl,binary_state):
 
     # compute with only one component present, other acts like ideal gas
     bmcsl.compute(state)
-    assert bmcsl.value == pytest.approx(10.*fex_cs(eta,d))
+    assert bmcsl.value == pytest.approx(volumes*fex_cs(eta,d))
```
mphoward, line 54: Just call the volume method:
```suggestion
    assert bmcsl.value == pytest.approx(state.mesh.full.volume()*fex_cs(eta,d))
```

python/tests/test_boublik_hard_sphere.py
```diff
@@ -60,7 +61,7 @@ def test_compute(bmcsl,binary_state):
     bmcsl.diameters['A'] = d
     bmcsl.diameters['B'] = d
     bmcsl.compute(state)
-    assert bmcsl.value == pytest.approx(10.*fex_cs(eta,d))
+    assert bmcsl.value == pytest.approx(volumes*fex_cs(eta,d))
```
mphoward, line 64: Please correct in all files, I'm going to stop marking these.

python/tests/test_crank_nicolson_integrator.py
```diff
@@ -51,8 +51,8 @@ def test_advance(state,grand,ig,linear,bd,cn):
 
     # add a linear field flux, but this is bulk so there should still be no change
     # except near the edges where the potential seems discontinuous by finite difference
-    linear.set_line('A', x=0., y=0., slope=0.25)
-    grand.external = linear
+    # linear.set_line('A', x=0., y=0., slope=0.25)
+    # grand.external = linear
```
mphoward, line 55: Don't comment stuff out from failing tests, or you will forget it needs to be fixed:
```suggestion
    linear.set_line('A', x=0., y=0., slope=0.25)
    grand.external = linear
```

python/tests/test_mesh.py
```diff
@@ -5,15 +5,18 @@
 
 @pytest.fixture
 def mesh():
-    return flyft.state.CartesianMesh(10.0,20,1)
+    #return flyft.state.CartesianMesh(10.0,20,1)
+    return flyft.state.SphericalMesh(10.0,20)
```
mphoward: Shouldn't this fixture be deleted now, and it will run on both defined in `conftest.py`?

python/tests/test_mesh.py
```diff
@@ -5,15 +5,18 @@
...
 def test_init(mesh):
     assert mesh.L == 10.0
     assert mesh.shape == 20
     assert mesh.step == pytest.approx(0.5)
     assert mesh.centers[0] == pytest.approx(0.25)
     assert mesh.centers[-1] == pytest.approx(9.75)
-
+    assert mesh.volume() == pytest.approx(4188.790204786391)#pytest.approx(10.)
+    assert mesh.volume(0) == pytest.approx(0.523598775598299)#pytest.approx(0.5)
+
```
mphoward, line 19: This looks like it belongs in a separate test for the `volume` method. You can define one test for the `cartesian_mesh` fixture and one for the `spherical_mesh` fixture so you can use different values for each

src/spherical_mesh.cc
```diff
@@ -31,15 +31,19 @@ std::shared_ptr<Mesh> SphericalMesh::slice(int start, int end) const
...
     {
-    const double r = center(i);
+    const double r = lower_bound(i);
     return 4.*M_PI*r*r;
     }
 
-
+double SphericalMesh::volume() const
+    {
+    return (4.*M_PI/3.)*(L_*L_*L_);
```
mphoward, line 40: This will fail if the domain is shifted (`origin_ != 0`). This should always work, I think:
```suggestion
    const double rlo = lower_bound(0);
    const double rhi = upper_bound(shape_-1);
    return (4.*M_PI/3.)*(rhi*rhi*rhi - rlo*rlo*rlo);
```

mphoward (changes requested):

python/tests/test_mesh.py
mayukh33: Yes! I made the change after pushing it on github 😅

python/tests/test_crank_nicolson_integrator.py
mayukh33: I forgot to uncomment this one! Thank you for finding it!

python/tests/test_brownian_diffusive_flux.py
```diff
@@ -40,32 +42,57 @@ def test_ideal(grand,ig,bd):
...
+        # flux is computed at the left edge
+        j = -2.0*(2.*np.pi/state.mesh.full.L)*np.cos(2*np.pi*(x-0.5*state.mesh.full.step)/state.mesh.full.L)
+        
+        k = bd.fluxes['A']
+        
+        assert np.allclose(bd.fluxes['A'], j, rtol=1e-3, atol=1e-3)
+    else:
+        raise Exception("Mesh not defined")
+
+def test_excess(grand,state_grand,ig,bd):
```
mphoward, line 91: Same idea for the testing here.

python/tests/test_brownian_diffusive_flux.py
```diff
@@ -84,28 +111,44 @@ def test_excess(grand,ig,bd):
...
+        raise Exception("Mesh not defined")
+        
 def test_external(state,grand,ig,walls,linear,bd):
     ig.volumes['A'] = 1.0
     grand.ideal = ig
     for w in walls:
         w.diameters['A'] = 0.0
     grand.external = flyft.external.CompositeExternalPotential(walls)
-
+    print(grand.external)
```
mphoward, line 151:
```suggestion
```

python/tests/test_crank_nicolson_integrator.py
```diff
@@ -70,12 +80,13 @@ def test_advance(state,grand,ig,linear,bd,cn):
...
     cn.advance(bd, grand, state, -1.5e-4)
-    assert state.time == pytest.approx(0.)
+    assert state.time == pytest.approx(0.)    
+
 
 @pytest.mark.parametrize("adapt",[False,True])
-def test_sine(adapt,cn):
-    state = flyft.State(2.,100,'A')
-    x = state.mesh.local.coordinates
+def test_sine(adapt,cn,state_sine):
```
mphoward, line 87: Same comment here, it's OK if you just want it to run on `CartesianMesh`.

mphoward (changes requested):

mphoward (approved):
Additional work is still needed, but we are merging into a staging branch to keep the PR manageable. Good job!

PR #36, opened 2023-04-20
Rosenfeld FMT for spherical geometry

mayukh33:


src/rosenfeld_fmt.cc
```diff
@@ -254,32 +235,499 @@ void RosenfeldFMT::_compute(std::shared_ptr<State> state, bool compute_value)
...
+                computeProportionalByWeight(w0, w1, wv1, w2, wv2, R);
+
+                dphi_dn0k_w0k(idx) = dphi_dn0k(idx)*w0;
+                dphi_dn1k_w1k(idx) = dphi_dn1k(idx)*w1;
+                dphi_dn2k_w2k(idx) = dphi_dn2k(idx)*w2;
+                dphi_dn3k_w3k(idx) = dphi_dn3k(idx)*w3;
+                // sign is opposite here due to oddness of weight function and reversed order
+                dphi_dnv1k_wv1k(idx) = -dphi_dnv1k(idx)*wv1;
+                dphi_dnv2k_wv2k(idx) = -dphi_dnv2k(idx)*wv2;
+                // TODO: there is a missing term for nv1 and nv2
```
mphoward, line 345: Remove TODO items when they are completed
```suggestion
```

src/rosenfeld_fmt.cc
```diff
@@ -254,32 +235,499 @@ void RosenfeldFMT::_compute(std::shared_ptr<State> state, bool compute_value)
...
+                dphi_dnv2k_w3k(idx) = -dphi_dnv2k(idx)*w3;
```
mphoward, line 346: Just want to check there should be a minus sign here because we are flipping the whole order of convolution (before we break it into two integrals) because w3 itself is even. I think it should be here, and we should add a comment as to why

src/rosenfeld_fmt.cc
```diff
@@ -254,32 +235,499 @@ void RosenfeldFMT::_compute(std::shared_ptr<State> state, bool compute_value)
...
+    // TODO: cache allocation
+    std::shared_ptr<Field> tmp_dF_dn0, tmp_dF_dn1, tmp_dF_dn2, tmp_dF_dn3, tmp_dF_dnv1, tmp_dF_dnv2, tmp_dF_dnv2_w3, tmp_dF_dnv1_w3;
+    setupField(tmp_dF_dn0);
+    setupField(tmp_dF_dn1);
+    setupField(tmp_dF_dn2);
+    setupField(tmp_dF_dn3);
+    setupField(tmp_dF_dnv1);
+    setupField(tmp_dF_dnv2);
+    setupField(tmp_dF_dnv2_w3);
+    setupField(tmp_dF_dnv1_w3);
```
mphoward, line 262: It might be helpful to denote the second term like `tmp_dF_dnv1_2` rather than `tmp_dF_dnv1_w3` to make clear it is the terms of the derivative?

src/rosenfeld_fmt.cc
```diff
@@ -254,32 +235,499 @@ void RosenfeldFMT::_compute(std::shared_ptr<State> state, bool compute_value)
...
+                const auto r = mesh->center(idx);
+                dF_dn0(idx) /= r;
+                dF_dn1(idx) /= r;
+                dF_dn2(idx) /= r;
+                dF_dn3(idx) /= r;
+                dF_dnv1(idx) /= r;
+                dF_dnv2(idx) /= r;
+                // TODO: extra term for nv1 and nv2
+                dF_dnv2(idx) = dF_dnv2_w3(idx)/(r*r)+dF_dnv2(idx);
+                dF_dnv1(idx) = dF_dnv1_w3(idx)/(r*r)+dF_dnv1(idx);
```
mphoward, line 406: Let's add a comment and use `+=` so it's obvious what this is doing
```suggestion
                // consolidate second term with first
                dF_dnv2(idx) += dF_dnv2_w3(idx)/(r*r);
                dF_dnv1(idx) += dF_dnv1_w3(idx)/(r*r);
```

src/rosenfeld_fmt.cc
```diff
@@ -254,32 +235,499 @@ void RosenfeldFMT::_compute(std::shared_ptr<State> state, bool compute_value)
...
+                }
+            }
+
+        // handle special case points near center
+        if (mesh->lower_bound(0) < R)
```
mphoward, line 411: I think I handled this correctly, but let's confirm that there is only one integral necessary here for the vector weighted terms. It is the last page of the notes. I think all we need to do is replace rho by dphi/dnv2.

src/rosenfeld_fmt.cc
```diff
@@ -254,32 +235,499 @@ void RosenfeldFMT::_compute(std::shared_ptr<State> state, bool compute_value)
...
+        }
+
+    // finish up communication of all types
+    for (const auto& t : state->getTypes())
+        {
+        state->getMesh()->endSync(derivatives_(t));
+        }
+    }
+
+void RosenfeldFMT::computeCartesianWeightedDensities(std::shared_ptr<State> state)
```
mphoward, line 501: Usually we like to put functions in the order they appear in the header. Making a note that we should align the `cc` file and `h` file function order at the very end. (Don't do it now or it will make a big change set that is just moving lines.)

src/rosenfeld_fmt.cc
```diff
@@ -254,32 +235,499 @@ void RosenfeldFMT::_compute(std::shared_ptr<State> state, bool compute_value)
...
+
+    // zero the weights before accumulating by type
+    std::fill(n0k_->view().begin(), n0k_->view().end(), 0.);
+    std::fill(n1k_->view().begin(), n1k_->view().end(), 0.);
+    std::fill(n2k_->view().begin(), n2k_->view().end(), 0.);
+    std::fill(n3k_->view().begin(), n3k_->view().end(), 0.);
+    std::fill(nv1k_->view().begin(), nv1k_->view().end(), 0.);
+    std::fill(nv2k_->view().begin(), nv2k_->view().end(), 0.);
+
+    for (const auto& t : state->getTypes())
```
mphoward, line 513: Making a note that we should think through whether we keep doing the Cartesian weights in this special way, or if we should do it more like the spherical ones. The only advantage of doing this is to reduce the number of Fourier transforms when there are like 1 or 2 types, and to reduce the communication overhead in general.

src/rosenfeld_fmt.cc
```diff
@@ -254,32 +235,499 @@ void RosenfeldFMT::_compute(std::shared_ptr<State> state, bool compute_value)
...
+    // zero the real space weighted densities for accumulation later
+    std::fill(n0_->view().begin(), n0_->view().end(), 0.);
+    std::fill(n1_->view().begin(), n1_->view().end(), 0.);
+    std::fill(n2_->view().begin(), n2_->view().end(), 0.);
+    std::fill(n3_->view().begin(), n3_->view().end(), 0.);
+    std::fill(nv1_->view().begin(), nv1_->view().end(), 0.);
+    std::fill(nv2_->view().begin(), nv2_->view().end(), 0.);
+
+    // weighted densities by type
+    // TODO: we should only be alloc'ing these once, will fix later
```
mphoward, line 597: Noting that this here should also be handled in the `tmp` allocation

src/rosenfeld_fmt.cc
```diff
@@ -254,32 +235,499 @@ void RosenfeldFMT::_compute(std::shared_ptr<State> state, bool compute_value)
...
+            continue;
+            }
+
+        // get the fourier transformed weighted densities for this type
+            {
+            fourierTransformFieldSpherical(state->getField(t)->const_full_view(), mesh);
+            auto rhok = ft_->const_view_reciprocal();
+            auto n2k = n2k_->view();
+            auto n3k = n3k_->view();
+            auto nv2k = nv2k_->view();
```
mphoward, line 622: These are candidates to become `tmp` variables as well

mphoward (commented):
Overall, I think this is in good shape! I noted some spots we should work on. Let's make a plan for who will implement these steps. Then, I think we are ready to test this code out.

We do still need to iron out the MPI communication for the spherical geometry, but that shouldn't get in the way of testing provided we pad the domain with enough zeros.

src/rosenfeld_fmt.cc
```diff
@@ -254,32 +235,499 @@ void RosenfeldFMT::_compute(std::shared_ptr<State> state, bool compute_value)
...
+        }
+    }
+
+void RosenfeldFMT::computeSphericalDerivative(std::shared_ptr<State> state)
+    {
+    const auto mesh = state->getMesh()->local().get();
+    const auto kmesh = ft_->getWavevectors();  
+
+    // these temporary variables will be used for convolutions of each derivative per type
+    // TODO: cache allocation
```
mphoward, line 253: Remove TODO comments as they are addressed
```suggestion
```

src/rosenfeld_fmt.cc
```diff
@@ -254,32 +235,499 @@ void RosenfeldFMT::_compute(std::shared_ptr<State> state, bool compute_value)
...
+    // zero the real space weighted densities for accumulation later
+    std::fill(n0_->view().begin(), n0_->view().end(), 0.);
+    std::fill(n1_->view().begin(), n1_->view().end(), 0.);
+    std::fill(n2_->view().begin(), n2_->view().end(), 0.);
+    std::fill(n3_->view().begin(), n3_->view().end(), 0.);
+    std::fill(nv1_->view().begin(), nv1_->view().end(), 0.);
+    std::fill(nv2_->view().begin(), nv2_->view().end(), 0.);
+
+    // weighted densities by type
+    // TODO: we should only be alloc'ing these once, will fix later
```
mphoward, line 597:
```suggestion
```

src/rosenfeld_fmt.cc
mphoward: These could use `tmp_complex_field_` rather than be member variables. You would want to delete the member variables.

src/rosenfeld_fmt.cc
```diff
@@ -254,32 +235,499 @@ void RosenfeldFMT::_compute(std::shared_ptr<State> state, bool compute_value)
...
+    {
+    const auto kmesh = ft_->getWavevectors();
+
+    // zero the weights before accumulating by type
+    std::fill(n0k_->view().begin(), n0k_->view().end(), 0.);
+    std::fill(n1k_->view().begin(), n1k_->view().end(), 0.);
+    std::fill(n2k_->view().begin(), n2k_->view().end(), 0.);
+    std::fill(n3k_->view().begin(), n3k_->view().end(), 0.);
+    std::fill(nv1k_->view().begin(), nv1k_->view().end(), 0.);
+    std::fill(nv2k_->view().begin(), nv2k_->view().end(), 0.);
```
mphoward, line 511: Same here for Cartesian, these should probably be `tmp_complex_field_` rather than member variables.

Unlike `n0_`, `n1_`, etc., these do not have meaning outside of their use in convolution. Therefore, they are `tmp`.

src/rosenfeld_fmt.cc
```diff
@@ -254,32 +235,499 @@ void RosenfeldFMT::_compute(std::shared_ptr<State> state, bool compute_value)
...
+
+    // weighted densities by type
+    // TODO: we should only be alloc'ing these once, will fix later
+    
+    setupField(tmp_field_["n0"]);
+    setupField(tmp_field_["n1"]);
+    setupField(tmp_field_["n2"]);
+    setupField(tmp_field_["n3"]);
+    setupField(tmp_field_["nv1"]);
+    setupField(tmp_field_["nv2"]);
```
mphoward, line 604: Are all 6 of these needed? It looks like only 3 are used below.

mphoward (commented):
You definitely have the right idea here! If the code continues to compile and run, this is exactly what I had in mind. Please take a look at the suggestions, then continue to address any comments that are simple fixes on this PR. Please resolve those from the Conversation tab as you do that (if they are trivial), or comment if you need to me to confirm they are done right.

src/rosenfeld_fmt.cc
mayukh33: It makes sense. I will change them to `tmp` variables.

src/mesh.cc
```diff
@@ -105,19 +105,33 @@ double Mesh::integrateVolume(int idx, const DataView<const double>& f) const
     return integrateVolume(idx, f(idx));
     }
 
-double Mesh::interpolate(int /*i*/,double f_lo, double f_hi) const
+double Mesh::interpolate(double x, const DataView<double>& f) const
     {
-    return 0.5*(f_lo+f_hi);
+    return interpolate(x,f);
```
mphoward, line 110: To confirm: this works and does not lead to infinite recursion?

src/rosenfeld_fmt.cc
```diff
@@ -434,35 +431,69 @@ void RosenfeldFMT::computeSphericalDerivative(std::shared_ptr<State> state)
...
-                for (int ig_idx=lower_bin; ig_idx < split_bin; ++ig_idx)
+                
+                double n = 100;
+                double x;
+                double dr;
+                
+                //Trapezoidal integral 1
+                dr = split/n;
+                x = lower;
+
```
mphoward, line 451: Even though `lower = 0`, you should still include it here in case you ever needed to change `lower` for some reason. This also makes the variable declaration more compact:
```suggestion
                // integral 1
                // we have found 100 points works well
                const int n = 100;
                double dr = (split-lower)/n;
                double x = lower;
```

src/rosenfeld_fmt.cc
```diff
@@ -434,35 +431,69 @@ void RosenfeldFMT::computeSphericalDerivative(std::shared_ptr<State> state)
...
+                double dF_dn3_1 = 0;
+                double dF_dn3_2 = 0;
                 // take integrals using quadrature
                 const auto r = mesh->center(idx);
-                const auto lower_bin = mesh->bin(0.);
-                const auto split_bin = mesh->bin(R-r);
-                const auto upper_bin = mesh->bin(r+R);
+                const auto lower = 0.;
+                const auto split = R-r;
+                const auto upper = r+R;
```
mphoward, line 441: I would be explicit about the type here:
```suggestion
                const double lower = 0.;
                const double split = R-r;
                const double upper = r+R;
```

src/rosenfeld_fmt.cc
```diff
@@ -434,35 +431,69 @@ void RosenfeldFMT::computeSphericalDerivative(std::shared_ptr<State> state)
...
+                    const double factor = (ig_idx == 0 || ig_idx == n) ? 0.5 : 1.0;
+                    dF_dn3_1 += factor * 4. * M_PI * dphi_dn3_ig * (x * x) ;
+                    x += dr;
                     }
-                for (int ig_idx=split_bin; ig_idx < upper_bin; ++ig_idx)
+                    
+                //Trapezoidal integral 2
+                dr = (upper-split)/n;
+                x = split;                
+
```
mphoward, line 463:
```suggestion
                // integral 2
                dr = (upper-split)/n;
                x = split;                
```

src/rosenfeld_fmt.cc
```diff
@@ -434,35 +431,69 @@ void RosenfeldFMT::computeSphericalDerivative(std::shared_ptr<State> state)
...
-                    dF_dnv1(idx) += dphi_dnv1(ig_idx) * kernel_wv1 * dr;
-                    dF_dnv2(idx) += dphi_dnv2(ig_idx) * kernel_wv2 * dr;
+                    
+                    const double factor = (ig_idx == 0 || ig_idx == n) ? 0.5 : 1.0;
+                    double dphi_dn0_ig = mesh->interpolate(x, dphi_dn0);
+                    double dphi_dn1_ig = mesh->interpolate(x, dphi_dn1);
+                    double dphi_dn2_ig = mesh->interpolate(x, dphi_dn2);
+                    double dphi_dn3_ig = mesh->interpolate(x, dphi_dn3);
+                    double dphi_dnv1_ig = mesh->interpolate(x, dphi_dnv1);
+                    double dphi_dnv2_ig = mesh->interpolate(x, dphi_dnv2);
```
mphoward, line 479: I would use `const auto` here, `const` since we know this will not change and `auto` since we want it to have the return type from `interpolate` (which will match the data type).
```suggestion
                    const auto dphi_dn0_ig = mesh->interpolate(x, dphi_dn0);
                    const auto dphi_dn1_ig = mesh->interpolate(x, dphi_dn1);
                    const auto dphi_dn2_ig = mesh->interpolate(x, dphi_dn2);
                    const auto dphi_dn3_ig = mesh->interpolate(x, dphi_dn3);
                    const auto dphi_dnv1_ig = mesh->interpolate(x, dphi_dnv1);
                    const auto dphi_dnv2_ig = mesh->interpolate(x, dphi_dnv2);
```

src/rosenfeld_fmt.cc
```diff
@@ -434,35 +431,69 @@ void RosenfeldFMT::computeSphericalDerivative(std::shared_ptr<State> state)
...
+                    dF_dn3_2 += factor * dphi_dn3_ig * kernel_w3 ;
+                    dF_dnv1(idx) += factor * dphi_dnv1_ig * kernel_wv1 ;
+                    dF_dnv2(idx) += factor * dphi_dnv2_ig* kernel_wv2 ;
+                    x += dr;               
                     }
+                
+                dF_dn0(idx) *= dr;
+                dF_dn1(idx) *= dr;
+                dF_dn2(idx) *= dr;
+                dF_dn3(idx) = (dF_dn3_1+dF_dn3_2)*dr;
```
mphoward, line 494: Don't these integrals have different step sizes? I would split this up: do `dF_dn3_1 *= dr;` after the first integral, then:
```suggestion
                dF_dn3(idx) = dF_dn3_1 + dF_dn3_2*dr; 
```
here.

src/rosenfeld_fmt.cc
```diff
@@ -669,36 +695,63 @@ void RosenfeldFMT::computeSphericalWeightedDensities(std::shared_ptr<State> stat
...
                     {
-                    const auto rig = mesh->center(ig_idx);
-                    n3i(idx) += (M_PI/r) * r_rho(ig_idx) * (R*R - (r-rig)*(r-rig)) * dr;
-                    n2i(idx) += (2.*M_PI*R/r) * r_rho(ig_idx) * dr;
-                    nv2i(idx) += (M_PI/(r*r)) * r_rho(ig_idx) * (R*R + r*r - rig*rig) * dr;
+                    const double factor = (ig_idx == 0 || ig_idx == n) ? 0.5 : 1.0;
+                    const auto rho_ig = mesh->interpolate(x,rho);
+                    n3_ig_2 += factor * (M_PI/r) * (rho_ig *x* (R*R - (r-x)*(r-x)));
+                    n2_ig_2 += factor * (2.*M_PI*R/r) * rho_ig * x;
+                    nv2_ig_2 += factor * (M_PI/(r*r)) * (rho_ig * x  * (R*R + r*r - x*x));
```
mphoward, line 748: Can't these be acculumated directly?
```suggestion
                    n2(idx) += factor * (2.*M_PI*R/r) * rho_ig * x;
                    nv2(idx) += factor * (M_PI/(r*r)) * (rho_ig * x  * (R*R + r*r - x*x));
```

src/rosenfeld_fmt.cc
```diff
@@ -669,36 +695,63 @@ void RosenfeldFMT::computeSphericalWeightedDensities(std::shared_ptr<State> stat
...
+            
+                //Initial step for the first integral 
+                double n3_ig_1 = 0;
+                double n3_ig_2 = 0; 
+                double n2_ig_2 = 0;
+                double nv2_ig_2 = 0;
+            
+                //Integral from 0 to R-r
+                x = lower;
+                dr = split/n;
```
mphoward, line 729:
```suggestion
                // integral 1
                const int n = 100;
                const double dr = (split-lower)/n;
                double x = lower;
```

src/rosenfeld_fmt.cc
```diff
@@ -669,36 +695,63 @@ void RosenfeldFMT::computeSphericalWeightedDensities(std::shared_ptr<State> stat
...
+                const double upper = r+R;
+                double x ;
+                double n = 100;
+                double dr;
+            
+                //Initial step for the first integral 
+                double n3_ig_1 = 0;
+                double n3_ig_2 = 0; 
+                double n2_ig_2 = 0;
+                double nv2_ig_2 = 0;
```
mphoward, line 725:
```suggestion
                double n3_ig_1 = 0;
                double n3_ig_2 = 0; 
```

src/rosenfeld_fmt.cc
```diff
@@ -669,36 +695,63 @@ void RosenfeldFMT::computeSphericalWeightedDensities(std::shared_ptr<State> stat
...
+                for (int ig_idx=0; ig_idx <=n; ig_idx++)
                     {
-                    n3i(idx) += 4. * M_PI * r_rho(ig_idx) * mesh->center(ig_idx) * dr;
+                    const double factor = (ig_idx == 0 || ig_idx == n) ? 0.5 : 1.0;
+                    double rho_ig = mesh->interpolate(x, rho);
+                    n3_ig_1 += factor * 4*M_PI* x * x * rho_ig;
+                    x += dr;
                     }
-                for (int ig_idx=split_bin; ig_idx < upper_bin; ++ig_idx)
+
```
mphoward, line 737: As in the functional derivative, make sure to scale `n3_ig_1 *= dr;` here.

src/rosenfeld_fmt.cc
```diff
@@ -669,36 +695,63 @@ void RosenfeldFMT::computeSphericalWeightedDensities(std::shared_ptr<State> stat
...
+                    const auto rho_ig = mesh->interpolate(x,rho);
+                    n3_ig_2 += factor * (M_PI/r) * (rho_ig *x* (R*R - (r-x)*(r-x)));
+                    n2_ig_2 += factor * (2.*M_PI*R/r) * rho_ig * x;
+                    nv2_ig_2 += factor * (M_PI/(r*r)) * (rho_ig * x  * (R*R + r*r - x*x));
+                    x +=dr;
                     }
+                    
+                n3i(idx) = (n3_ig_1+ n3_ig_2)*dr;
+                n2i(idx) = n2_ig_2*dr;
+                nv2i(idx) = nv2_ig_2*dr;
```
mphoward, line 754:
```suggestion
                n3i(idx) = n3_ig_1 + n3_ig_2*dr;
                n2i(idx) *= dr;
                nv2i(idx) *= dr;
```

src/rosenfeld_fmt.cc
```diff
@@ -726,6 +779,8 @@ void RosenfeldFMT::computeSphericalWeightedDensities(std::shared_ptr<State> stat
                 n3(idx) += n3i(idx);
                 nv1(idx) += nv1i;
                 nv2(idx) += nv2i(idx);
+                
+                std::cout<<mesh->center(idx)<<", "<<n3(idx)<<", "<<n2(idx) <<", "<<nv2(idx)<<std::endl;
```
mphoward, line 783: Don't commit debug statements.
```suggestion
```

mphoward (changes requested):
Overall this looks it should work! I had some suggestions, and one or two small issues of correctness. Please address those.

I didn't carefully check the integral formulas, so if you want to do that please let me know to have a look at this in a final review before merging..

src/mesh.cc
mayukh33: Yes! It works!

mayukh33: Yes! It works!

src/rosenfeld_fmt.cc
mphoward: Will not be handled in this PR.

mphoward (approved):
Some todo items (fix interpolation scheme and boundary conditions), will be handled in subsequent PRs!

PR #37, opened 2023-07-18
Add different boundary types

mayukh33:


mphoward (commented):
Here are some initial thoughts, let me know if you want to discuss!

include/flyft/mesh.h
```diff
@@ -72,12 +71,19 @@ class Mesh
         
         bool operator==(const Mesh& other) const;
         bool operator!=(const Mesh& other) const;
+        
+        BoundaryType getlower_boundary_type() const;
+        BoundaryType getupper_boundary_type() const;
```
mphoward, line 76: These should be added after the other getters, after line 51. For consistency with other methods in this class, they should be named like:
```suggestion
        BoundaryType lower_boundary_condition() const;
        BoundaryType upper_boundary_condition() const; 
```
We can revisit in a later PR whether we want to unify the naming convention for methods across classes to use more conventional get/set.

python/flyft/_flyft/_flyft.h
```diff
@@ -2,6 +2,7 @@
 #define FLYFT_PYTHON__FLYFT_H_
 
 // flyft headers for binding opaque types
+#include "flyft/boundary_type.h"
```
mphoward, line 5: I don't think this include is needed here, since it is not used below.

include/flyft/mesh.h
```diff
@@ -1,6 +1,7 @@
 #ifndef FLYFT_MESH_H_
 #define FLYFT_MESH_H_
 #include "flyft/data_view.h"
+#include "flyft/boundary_type.h"
```
mphoward, line 4: Add headers alphabetically
```suggestion
#include "flyft/boundary_type.h"
#include "flyft/data_view.h"
```

include/flyft/mesh.h
```diff
@@ -72,12 +71,19 @@ class Mesh
...
+        BoundaryType getlower_boundary_type() const;
+        BoundaryType getupper_boundary_type() const; 
 
     protected:
+        double lower_;    
+        double upper_;
         double L_;      //!< Length of the domain
         int shape_;     //!< Shape of the mesh
         double step_;   //!< Spacing between mesh points
         double origin_; //!< Origin of the mesh
```
mphoward, line 84: We no longer need `origin_` if we have `lower_`, since they are the same. You should remove the accessor `origin()`. Then, provide overloads of `lower_bound()` and `upper_bound()` that take no argument to give the bounds of the whole Mesh (rather than a specific bin), like we did for `volume()`.

Also, should we call these `lower_bound_` and `upper_bound_` internally to make it clearer what they represent?

python/flyft/_flyft/boundary_type.cc
```diff
@@ -0,0 +1,11 @@
+#include "_flyft.h"
+#include "flyft/boundary_type.h"
+
+#include <pybind11/stl_bind.h>
+
+void bindBoundaryType(py::module_& m)
+    {
+    using namespace flyft;
+
+    py::bind_map<BoundaryType>(m, "BoundaryType");
```
mphoward, line 10: This isn't the right way to bind this, you need to use `py::enum_`:
https://pybind11.readthedocs.io/en/stable/classes.html#enumerations-and-internal-types
I believe you should *not* call `export_values()` so that this stays scoped in `BoundaryType`.

src/cartesian_mesh.cc
```diff
@@ -1,31 +1,17 @@
 #include "flyft/cartesian_mesh.h"
+#include <iostream>
```
mphoward, line 2: Why this include? Use `<exception>` for exceptions.

src/cartesian_mesh.cc
```diff
@@ -1,31 +1,17 @@
...
-    : Mesh(shape,step,origin),area_(area)
+CartesianMesh::CartesianMesh(double lower,double upper,int shape, BoundaryType lower_bc, BoundaryType upper_bc,double area)
+    : Mesh(lower,upper,shape,lower_bc,upper_bc), area_(area)
     {
     }
 
 std::shared_ptr<Mesh> CartesianMesh::slice(int start, int end) const
     {
-    return std::shared_ptr<Mesh>(new CartesianMesh(end-start,step_,area_,lower_bound(start)));
+    return std::shared_ptr<Mesh>(new CartesianMesh(lower_bound(start),end,end-start,lower_bc_, upper_bc_,area_));
```
mphoward, line 14: This is not quite right. It also needs to account for the possible change in boundary conditions on slicing. I think this will work:
```suggestion
    return std::shared_ptr<Mesh>(new CartesianMesh(
        lower_bound(start),
        lower_bound(end),
        end-start,
        (start > 0) ? BoundaryType::internal : lower_bc_,
        (end < shape_-1) ? BoundaryType::internal : upper_bc_,
        area_
    ));
```

src/cartesian_mesh.cc
```diff
@@ -48,4 +34,17 @@ double CartesianMesh::gradient(int /*i*/, double f_lo, double f_hi) const
...
+void CartesianMesh::validateBoundaryCondition(){
+    if (lower_bc_ != BoundaryType::periodic && upper_bc_ == BoundaryType::periodic){
+        throw std::invalid_argument("Lower boundary type not valid for periodic boundary condition");
+        }
+    else if (lower_bc_ == BoundaryType::periodic && upper_bc_ != BoundaryType::periodic){
+        throw std::invalid_argument("Upper boundary type not valid for periodic boundary condition");
+        }
+    else{
+        return;
+        }
```
mphoward, line 46: You don't need an `else` branch that does nothing.
```suggestion
```

src/cartesian_mesh.cc
```diff
@@ -48,4 +34,17 @@ double CartesianMesh::gradient(int /*i*/, double f_lo, double f_hi) const
     return (f_hi-f_lo)/step_; 
     }
 
+void CartesianMesh::validateBoundaryCondition(){
+    if (lower_bc_ != BoundaryType::periodic && upper_bc_ == BoundaryType::periodic){
+        throw std::invalid_argument("Lower boundary type not valid for periodic boundary condition");
+        }
+    else if (lower_bc_ == BoundaryType::periodic && upper_bc_ != BoundaryType::periodic){
+        throw std::invalid_argument("Upper boundary type not valid for periodic boundary condition");
+        }
```
mphoward, line 43: This is going to cause issues when slicing, since you can get `internal` + `periodic`. You should first check if either boundary condition is internal and `return` if true since you can't do any meaningful additional checks. Then, you can combine this into one check like:

```suggestion
    if ((lower_bc_ != BoundaryType::periodic && upper_bc_ == BoundaryType::periodic) ||
        (lower_bc_ == BoundaryType::periodic && upper_bc_ != BoundaryType::periodic))
        {
        throw std::invalid_argument("Both boundaries must be periodic if one is.");
        }
```

src/fourier_transform.cc
```diff
@@ -7,26 +7,23 @@
...
     {
     #ifdef FLYFT_OPENMP
     // use all available OpenMP threads
     fftw_init_threads();
     fftw_plan_with_nthreads(omp_get_max_threads());
     #endif
-
     // this is the doc'd size of "real" memory required for the r2c / c2r transform
-    data_ = fftw_alloc_real(2*(mesh_.shape()/2+1));
+    data_ = fftw_alloc_real(2*(N/2+1));
```
mphoward, line 19:
```suggestion
    data_ = fftw_alloc_real(2*(N_/2+1));
```

src/fourier_transform.cc
```diff
@@ -7,26 +7,23 @@
...
 
-    r2c_plan_ = fftw_plan_dft_r2c_1d(mesh_.shape(),
+    r2c_plan_ = fftw_plan_dft_r2c_1d(N,
```
mphoward, line 21:
```suggestion
    r2c_plan_ = fftw_plan_dft_r2c_1d(N_,
```

src/fourier_transform.cc
```diff
@@ -7,26 +7,23 @@
...
                                      data_,
                                      reinterpret_cast<fftw_complex*>(data_),
                                      FFTW_ESTIMATE);
 
-    c2r_plan_ = fftw_plan_dft_c2r_1d(mesh_.shape(),
+    c2r_plan_ = fftw_plan_dft_c2r_1d(N,
```
mphoward, line 26:
```suggestion
    c2r_plan_ = fftw_plan_dft_c2r_1d(N_,
```

src/fourier_transform.cc
```diff
@@ -119,15 +116,21 @@ void FourierTransform::transform()
         {
         // execute inverse FFT and renormalize by N (FFTW does not)
         fftw_execute(c2r_plan_);
-        const auto size = mesh_.shape();
+        const auto size = getN();
         std::transform(data_,data_+size,data_,[&](auto x){return x/size;});
```
mphoward, line 120:
```suggestion
        std::transform(data_,data_+size,data_,[&](auto x){return x/N_;});
```

src/mesh.cc
```diff
@@ -1,38 +1,24 @@
 #include "flyft/mesh.h"
+#include <iostream>
```
mphoward, line 2: This header is not used
```suggestion
```

src/mesh.cc
```diff
@@ -57,7 +43,7 @@ int Mesh::shape() const
 
 double Mesh::step() const
     {
-    return step_;
+    return ((upper_-lower_)/shape_);
```
mphoward, line 46: Revert this:
```suggestion
    return step_;
```

src/mesh.cc
```diff
@@ -72,7 +58,7 @@ int Mesh::asShape(double dx) const
 
 double Mesh::asLength(int shape) const
     {
-    return shape*step_;
+    return shape*step();
```
mphoward, line 61: Revert this:
```suggestion
    return shape*step_;
```

src/spherical_mesh.cc
```diff
@@ -1,33 +1,19 @@
...
-
-SphericalMesh::SphericalMesh(int shape, double step, double origin)
-    : Mesh(shape, step, origin)
-    {
-    }
 
 std::shared_ptr<Mesh> SphericalMesh::slice(int start, int end) const
     {
-    return std::shared_ptr<Mesh>(new SphericalMesh(end-start,step_,lower_bound(start)));
+    return std::shared_ptr<Mesh>(new SphericalMesh(lower_bound(start),end,end-start,lower_bc_, upper_bc_));
```
mphoward, line 16: Fix slicing logic here like in `CartesianMesh`

src/spherical_mesh.cc
```diff
@@ -59,5 +45,16 @@ double SphericalMesh::gradient(int idx, double f_lo, double f_hi) const
...
+    {
+    if (lower_bc_ == BoundaryType::periodic){
+        throw std::invalid_argument("Lower boundary type not valid for spherical geometry");
+        }
+    else if (upper_bc_ == BoundaryType::periodic){
+        throw std::invalid_argument("Upper boundary type not valid for spherical geometry");
+        }
+    else{
+        return;
+        }
```
mphoward, line 58:
```suggestion
    if (lower_bc_ == BoundaryType::periodic || upper_bc_ == BoundaryType::periodic){
        throw std::invalid_argument("Periodic boundary conditions not valid in spherical geometry");
        }
```

mphoward (changes requested):
This is a good first step! Here are some comments

src/cartesian_mesh.cc
```diff
@@ -48,4 +34,13 @@ double CartesianMesh::gradient(int /*i*/, double f_lo, double f_hi) const
     return (f_hi-f_lo)/step_; 
     }
 
+void CartesianMesh::validateBoundaryCondition(){
```
mphoward, line 37: I think we should add a `virtual void Mesh::validateBoundaryCondition` that always does this check. There is no geometry where this would be permitted. Then, deriving classes can override it if additional checks are needed.

CartesianMesh would not need to override it.

src/fourier_transform.cc
```diff
@@ -119,15 +116,21 @@ void FourierTransform::transform()
         {
         // execute inverse FFT and renormalize by N (FFTW does not)
         fftw_execute(c2r_plan_);
-        const auto size = mesh_.shape();
-        std::transform(data_,data_+size,data_,[&](auto x){return x/size;});
+        const auto size = getN();
+        std::transform(data_,data_+size,data_,[&](auto x){return x/N_;});
```
mphoward, line 120:
```suggestion
        std::transform(data_,data_+N_,data_,[&](auto x){return x/N_;});
```

src/mesh.cc
```diff
@@ -146,12 +136,23 @@ double Mesh::gradient(int idx, const DataView<const double>& f) const
 
 bool Mesh::operator==(const Mesh& other) const
```
mphoward, line 137: Can you grep the code to see where we are actually doing an `==` check on a Mesh? I am wondering if this can return true even when they do not actually match, e.g., they are different types.

src/spherical_mesh.cc
```diff
@@ -52,12 +38,17 @@ double SphericalMesh::volume(int i) const
...
+    if (idx == 0 && lower_ == 0.){
         return 0;
         }
     else{
         return (f_hi-f_lo)/(step_);
         }
     }   
-
+void SphericalMesh::validateBoundaryCondition()
+    {
```
mphoward, line 49: After making the `Mesh` implementation, you can call it first here then do the additional checks:
```suggestion
    {
    Mesh::validateBoundaryConditions();
```

mphoward (changes requested):

src/fourier_transform.cc
mphoward: N_ should be an int. Maybe you accidentally made it a double when you declared it in the header? I might have missed  that.

python/tests/test_mesh.py
```diff
@@ -25,9 +32,12 @@ def test_immutable(mesh):
         mesh.step = 1.0
 
 def test_volume_cartesian(cartesian_mesh):
+    assert cartesian_mesh.lower_boundary_condition ==  cartesian_mesh.boundary_type.periodic
+    assert cartesian_mesh.upper_boundary_condition ==  cartesian_mesh.boundary_type.periodic
```
mphoward, line 36: Once we change to strings, these can be:
```suggestion
    assert cartesian_mesh.lower_boundary_condition ==  "periodic"
    assert cartesian_mesh.upper_boundary_condition ==  "periodic"
```

python/tests/test_mesh.py
```diff
@@ -25,9 +32,12 @@ def test_immutable(mesh):
...
     assert cartesian_mesh.volume() == pytest.approx(10.)
     assert cartesian_mesh.volume(0) == pytest.approx(0.5)
 
 def test_volume_spherical(spherical_mesh):
+    assert spherical_mesh.lower_boundary_condition ==  spherical_mesh.boundary_type.reflect
```
mphoward, line 41:
```suggestion
    assert spherical_mesh.lower_boundary_condition ==  "reflect"
```
You also need to test the upper bc.

src/cartesian_mesh.cc
```diff
@@ -3,29 +3,14 @@
...
-    : Mesh(shape,step,origin),area_(area)
+CartesianMesh::CartesianMesh(double lower,double upper,int shape, BoundaryType lower_bc, BoundaryType upper_bc,double area)
+    : Mesh(lower,upper,shape,lower_bc,upper_bc), area_(area)
     {
     }
 
 std::shared_ptr<Mesh> CartesianMesh::slice(int start, int end) const
     {
-    return std::shared_ptr<Mesh>(new CartesianMesh(end-start,step_,area_,lower_bound(start)));
+    return std::shared_ptr<Mesh>(new CartesianMesh(lower_bound(start),lower_bound(end),end-start, (start > 0)?BoundaryType::internal:lower_bc_, (start > 0)?BoundaryType::internal:upper_bc_,area_));
```
mphoward, line 13: This should use the end for the upper BC, and also needs better formatting like this:
```suggestion
    return std::shared_ptr<Mesh>(new CartesianMesh(
        lower_bound(start),
        lower_bound(end),
        end-start,
        (start > 0) ? BoundaryType::internal : lower_bc_,
        (end < shape_) ? BoundaryType::internal : upper_bc_,
        area_));
```
We should double check that the way we have defined end means it should be less than `shape_`, I think that's right.

src/cartesian_mesh.cc
```diff
@@ -47,5 +32,8 @@ double CartesianMesh::gradient(int /*i*/, double f_lo, double f_hi) const
     {
     return (f_hi-f_lo)/step_; 
     }
-
 }
+
+
+
+
```
mphoward, line 39:
```suggestion
```

src/fourier_transform.cc
```diff
@@ -7,26 +7,23 @@
 
 namespace flyft
 {
-FourierTransform::FourierTransform(double L, int shape)
-    : mesh_(L,shape,-0.5*L/shape),
-      kmesh_(L,shape),
-      space_(RealSpace)
+FourierTransform::FourierTransform(double L, int N)
+    : L_(L), N_(N), kmesh_(L,N),space_(RealSpace)
```
mphoward, line 11:
```suggestion
    : L_(L), N_(N), kmesh_(L,N), space_(RealSpace)
```

src/fourier_transform.cc
```diff
@@ -7,26 +7,23 @@
...
     {
     #ifdef FLYFT_OPENMP
     // use all available OpenMP threads
     fftw_init_threads();
     fftw_plan_with_nthreads(omp_get_max_threads());
     #endif
-
     // this is the doc'd size of "real" memory required for the r2c / c2r transform
```
mphoward, line 18:
```suggestion

    // this is the doc'd size of "real" memory required for the r2c / c2r transform
```

src/fourier_transform.cc
```diff
@@ -119,15 +116,20 @@ void FourierTransform::transform()
...
         space_ = RealSpace;
         }
     }
 
-const Mesh& FourierTransform::getMesh() const
+double FourierTransform::getL() const
+    {
+    return L_;
+    }
+
```
mphoward, line 128:
```suggestion
```

src/mesh.cc
```diff
@@ -5,44 +5,41 @@
 namespace flyft
 {
 
-Mesh::Mesh(double L, int shape)
-    : Mesh(L,shape,0)
+Mesh::Mesh(double lower, double upper, int shape, BoundaryType lower_bc, BoundaryType upper_bc)
+    : lower_(lower), upper_(upper), L_(upper_-lower_), shape_(shape), step_(L_/shape_), lower_bc_(lower_bc), upper_bc_(upper_bc)
```
mphoward, line 9: You will need to reorder these after you change the order of them in the header, to avoid a compiler error.

src/mesh.cc
```diff
@@ -5,44 +5,41 @@
...
     {
+        validateBoundaryCondition();
```
mphoward, line 11:
```suggestion
    validateBoundaryCondition();
```

src/mesh.cc
```diff
@@ -5,44 +5,41 @@
...
+    return static_cast<int>((x-lower_)/step_);
     }
 
-Mesh::Mesh(int shape, double step, double origin)
-    : L_(shape*step), shape_(shape), step_(step), origin_(origin)
-    {
-    }
 
-double Mesh::center(int i) const
+double Mesh::lower_bound() const
```
mphoward, line 25: Reorder functions in this file based on the header.

src/mesh.cc
mphoward: This still needs to be looked into. This function either needs to be rewritten or removed.

src/parallel_mesh.cc
```diff
@@ -150,40 +150,107 @@ void ParallelMesh::startSync(std::shared_ptr<Field> field)
         // nothing to do, no buffer needed
         return;
         }
-
+
```
mphoward, line 153:
```suggestion

```

src/rosenfeld_fmt.cc
```diff
@@ -893,11 +893,12 @@ bool RosenfeldFMT::setup(std::shared_ptr<State> state, bool compute_value)
...
-    if (!ft_ || ft_mesh != ft_->getMesh())
-        {
-        ft_ = std::make_unique<FourierTransform>(ft_mesh.L(),ft_mesh.shape());
-        }
+    const double buffered_L = mesh->asLength(buffered_shape);
+    if (!ft_ || buffered_L != ft_->getL() || buffered_shape != ft_->getN())
+    {
+        ft_ = std::make_unique<FourierTransform>(buffered_L,buffered_shape);
+    }
+
```
mphoward, line 901:
```suggestion
        {
        ft_ = std::make_unique<FourierTransform>(buffered_L,buffered_shape);
        }
```

src/spherical_mesh.cc
```diff
@@ -1,33 +1,20 @@
 #include "flyft/spherical_mesh.h"
 
 #include <cmath>
+#include <iostream>
```
mphoward, line 4:
```suggestion
#include <exception>
```

src/spherical_mesh.cc
```diff
@@ -1,33 +1,20 @@
...
 
 namespace flyft
 {
-
-SphericalMesh::SphericalMesh(double R, int shape)
-    : SphericalMesh(R,shape,0)
+SphericalMesh::SphericalMesh(double lower, double upper,int shape, BoundaryType lower_bc, BoundaryType upper_bc)
+    : Mesh(lower, upper, shape, lower_bc, upper_bc)
     {
+          validateBoundaryCondition();
```
mphoward, line 11:
```suggestion
      validateBoundaryCondition();
```

src/spherical_mesh.cc
```diff
@@ -52,12 +39,17 @@ double SphericalMesh::volume(int i) const
 
 double SphericalMesh::gradient(int idx, double f_lo, double f_hi) const
     {
-    if (idx == 0 && origin_ == 0.){
+    if (idx == 0 && lower_ == 0.){
```
mphoward, line 42: I think we can do this based on the BCs now. Actually, this is true for ALL of the classes, so move it to base Mesh. It also is true of the upper BC so it should be
```suggestion
    if ((idx == 0 && lower_bc_ == BoundaryType::reflect) ||
        (idx == shape_-1 && upper_bc_ == BoundaryType::reflect)
        {
```

So, move this to `Mesh` please!

src/spherical_mesh.cc
mphoward: This still should be done here to make it explicit that this check is being called, see below.

src/spherical_mesh.cc
```diff
@@ -52,12 +39,17 @@ double SphericalMesh::volume(int i) const
...
     else{
         return (f_hi-f_lo)/(step_);
         }
     }   
-
+void SphericalMesh::validateBoundaryCondition()
+    {
+    if(lower_bc_ == BoundaryType::periodic || upper_bc_ == BoundaryType::periodic){
+         std::cout<<"Periodic boundary conditions not valid in spherical geometry";
+        }
```
mphoward, line 53:
```suggestion
    Mesh::validateBoundaryConditions();
    if(lower_bc_ == BoundaryType::periodic || upper_bc_ == BoundaryType::periodic)
        {
         throw std::invalid_argument("Periodic boundary conditions not valid in spherical geometry");
        }
```

mphoward: We also need to reject `upper_bc_ == BoundaryType::reflect`.

include/flyft/parallel_mesh.h
```diff
@@ -24,7 +24,6 @@ class ParallelMesh
 
         std::shared_ptr<const Mesh> local() const;
         std::shared_ptr<const Mesh> full() const;
-
         int getProcessorShape() const;
```
mphoward, line 27:
```suggestion

        int getProcessorShape() const;
```

python/flyft/_flyft/mesh.cc
```diff
@@ -1,5 +1,6 @@
 #include "_flyft.h"
 #include "flyft/mesh.h"
+#include "flyft/boundary_type.h"
```
mphoward, line 3:
```suggestion
#include "flyft/boundary_type.h"
#include "flyft/mesh.h"
```

python/flyft/state.py
```diff
@@ -59,19 +59,59 @@ def volume(self, bin = None):
...
+    
+    def upper_bound(self, bin=None):
+        if bin is None:
+            u = self._self.upper_bound
+        else:
+            u = self._self.bin_upper_bound(bin)
+        return u
+    
+    @staticmethod
+    def _parse_boundary_condition(bc):
```
mphoward, line 78: The changes we made to `_parse_boundary_condition` have not been committed, please add them.

python/flyft/state.py
```diff
@@ -59,19 +59,59 @@ def volume(self, bin = None):
...
-    origin = mirror.Property()
     shape = mirror.Property()
     step = mirror.Property()
 
 class CartesianMesh(Mesh,mirrorclass=_flyft.CartesianMesh):
-    def __init__(self,L,shape,area = 1.):      
-        super().__init__(L,shape,area) 
-
+    def __init__(self, L, shape, boundary_condition, area = 1.): 
+        bcs = Mesh._parse_boundary_condition(boundary_condition)
```
mphoward, line 106: The changes we discussed also need to be applied here.

python/tests/conftest.py
```diff
@@ -16,11 +16,13 @@ def grand():
 
 @pytest.fixture
 def cartesian_mesh_sine():
-    return flyft.state.CartesianMesh(2.,100,1)
+    return flyft.state.CartesianMesh(2.,100,"periodic",1)
+
```
mphoward, line 20:
```suggestion
```

python/tests/conftest.py
```diff
@@ -16,11 +16,13 @@ def grand():
...
 def cartesian_mesh_sine():
-    return flyft.state.CartesianMesh(2.,100,1)
+    return flyft.state.CartesianMesh(2.,100,"periodic",1)
+    
 
 @pytest.fixture
 def spherical_mesh_sine():
-    return flyft.state.SphericalMesh(2.,100)
+    return flyft.state.SphericalMesh(2.,100,"zero")
+
```
mphoward, line 25:
```suggestion
```

python/tests/test_mesh.py
```diff
@@ -6,28 +6,33 @@
...
     with pytest.raises(AttributeError):
         mesh.L = 20.0
     with pytest.raises(AttributeError):
-        mesh.shape = 10
+        mesh.shape = 100
     with pytest.raises(AttributeError):
         mesh.step = 1.0
 
 def test_volume_cartesian(cartesian_mesh):
+    assert cartesian_mesh.lower_boundary_condition ==  "periodic"
```
mphoward, line 30: These tests will not pass without the correct `_parse_boundary_conditions`

src/mesh.cc
mphoward: Still waiting for a comment back on this @mayukh33!

src/spherical_mesh.cc
```diff
@@ -1,33 +1,24 @@
...
 
 namespace flyft
 {
-
-SphericalMesh::SphericalMesh(double R, int shape)
-    : SphericalMesh(R,shape,0)
+SphericalMesh::SphericalMesh(double lower_bound, double upper_bound,int shape, BoundaryType lower_bc, BoundaryType upper_bc)
+    : Mesh(lower_bound, upper_bound, shape, lower_bc, upper_bc)
     {
+      validateBoundaryCondition();
```
mphoward, line 11:
```suggestion
    validateBoundaryCondition();
```

src/spherical_mesh.cc
```diff
@@ -1,33 +1,24 @@
...
-    }
 
 std::shared_ptr<Mesh> SphericalMesh::slice(int start, int end) const
     {
-    return std::shared_ptr<Mesh>(new SphericalMesh(end-start,step_,lower_bound(start)));
+    return std::shared_ptr<Mesh>(new SphericalMesh(
+        lower_bound(start),
+        lower_bound(end),end-start, 
+        (start > 0)?BoundaryType::internal:lower_bc_, 
+        (start > 0)?BoundaryType::internal:upper_bc_));
```
mphoward, line 21: This change did not get applied correctly (importantly, the second ternary check was wrong):
```suggestion
        lower_bound(start),
        lower_bound(end),
        end-start, 
        (start > 0) ? BoundaryType::internal : lower_bc_, 
        (end > shape_) ? BoundaryType::internal : upper_bc_));
```

src/spherical_mesh.cc
```diff
@@ -50,14 +41,18 @@ double SphericalMesh::volume(int i) const
...
-
+    
+void SphericalMesh::validateBoundaryCondition() const
+    {
+    Mesh::validateBoundaryCondition();
+    
+    if(lower_bc_ == BoundaryType::periodic || upper_bc_ == BoundaryType::periodic || upper_bc_ == BoundaryType::reflect)
+        {
+         throw std::invalid_argument("Periodic boundary conditions or reflect upper boundary condition is  not valid in spherical geometry");
+        }
```
mphoward, line 56: Make these separate so the message can be more specific
```suggestion
    if(lower_bc_ == BoundaryType::periodic || upper_bc_ == BoundaryType::periodic)
        {
         throw std::invalid_argument("Periodic boundary conditions invalid in spherical geometry");
        }
    else if  (upper_bc_ == BoundaryType::reflect)
        {
        throw std::invalid_argument("Reflect boundary condition invalid for upper boundary");
        }
```

mphoward (changes requested):
This PR is almost there! However, I think the changes we made in `state.py` did not get committed, so the unit tests will not pass. I am going to batch commit some small changes via GitHub so please pull then correct this so we can merge.

mphoward (approved):
Looks great, thank you!

PR #40, opened 2023-08-18
Add RPY diffusive flux

mayukh33:


include/flyft/brownian_diffusive_flux.h
```diff
@@ -10,11 +10,16 @@
...
+class BrownianDiffusiveFlux : public Flux
     {
     public:
+        TypeMap<double>& getDiffusivities();
+        const TypeMap<double>& getDiffusivities() const;
         void compute(std::shared_ptr<GrandPotential> grand, std::shared_ptr<State> state) override;
         int determineBufferShape(std::shared_ptr<State> state, const std::string& type) override;
+        
+    protected:
+        TypeMap<double> diffusivities_;
```
mphoward, line 22: I usually like getters to come after the "main" method. The diffusivities can also be private since we don't anticipate deriving from this class.
```suggestion
        void compute(std::shared_ptr<GrandPotential> grand, std::shared_ptr<State> state) override;

        TypeMap<double>& getDiffusivities();
        const TypeMap<double>& getDiffusivities() const;

        int determineBufferShape(std::shared_ptr<State> state, const std::string& type) override;
        
    private:
        TypeMap<double> diffusivities_;
```

include/flyft/rpy_diffusive_flux.h
```diff
@@ -0,0 +1,32 @@
...
+#define FLYFT_RPY_DIFFUSIVE_FLUX_H_
+
+#include "flyft/flux.h"
+#include "flyft/grand_potential.h"
+#include "flyft/state.h"
+#include "flyft/type_map.h"
+
+#include <memory>
+#include <exception>
+#include <bits/stdc++.h>
```
mphoward, line 11: Don't include headers where you don't use them. If you need something in the `.cc` file, put it there.
```suggestion
```

include/flyft/rpy_diffusive_flux.h
```diff
@@ -0,0 +1,32 @@
...
+class RPYDiffusiveFlux : public Flux
+    {
+    public:
+    
+        TypeMap<double>& getDiameters();
+        const TypeMap<double>& getDiameters() const;
+        void compute(std::shared_ptr<GrandPotential> grand, std::shared_ptr<State> state) override;
+        int determineBufferShape(std::shared_ptr<State> state, const std::string& type) override;
+        double getViscosity();
+        double setViscosity(double viscosity);
```
mphoward, line 25: Reordering these. Also note that the getter for a regular value should be `const` (since it cannot modify the class), and the setter for a value should return void.
```suggestion
        void compute(std::shared_ptr<GrandPotential> grand, std::shared_ptr<State> state) override;

        TypeMap<double>& getDiameters();
        const TypeMap<double>& getDiameters() const;

        double getViscosity() const;
        void setViscosity(double viscosity);

        int determineBufferShape(std::shared_ptr<State> state, const std::string& type) override;
```

python/flyft/_flyft/_flyft.cc
```diff
@@ -41,6 +41,7 @@ void bindFlux(py::module_&);
 void bindCompositeFlux(py::module_&);
 void bindDiffusiveFlux(py::module_&);
```
mphoward, line 42: Don't declare a function that is not used, this will lead to a link error.
```suggestion
```

python/flyft/_flyft/brownian_diffusive_flux.cc
```diff
@@ -3,11 +3,29 @@
...
+        using BrownianDiffusiveFlux::BrownianDiffusiveFlux;
+
+        //! pybind11 override of pure virtual compute method
+        void compute(std::shared_ptr<GrandPotential> grand, std::shared_ptr<State> state) override
+            {
+            PYBIND11_OVERRIDE_PURE(void, BrownianDiffusiveFlux, compute, grand, state);
+            }
+    };
+}
+
```
mphoward, line 22: This should not be needed, since this class has no pure virtual methods. The trampoline on `Flux` should handle that.

python/flyft/_flyft/rpy_diffusive_flux.cc
```diff
@@ -0,0 +1,13 @@
...
+#include "flyft/rpy_diffusive_flux.h"
+
+#include <pybind11/stl.h>
+
+void bindRPYDiffusiveFlux(py::module_& m)
+    {
+    using namespace flyft;
+
+    py::class_<RPYDiffusiveFlux,std::shared_ptr<RPYDiffusiveFlux>,Flux>(m, "RPYDiffusiveFlux")
+        .def(py::init<>())
```
mphoward, line 11: You'll need to expose the diameters like you did the diffusivities, and then the viscosity as a property here.

src/rpy_diffusive_flux.cc
```diff
@@ -0,0 +1,149 @@
...
+        }
+    if (external)
+        {
+        external->compute(state,false);
+        }
+    
+    
+    // compute fluxes on the left edge of the volumes (exclude the first point)
+    const auto mesh = state->getMesh()->local().get();
+    if (dynamic_cast<const SphericalMesh*>(mesh) != nullptr)
```
mphoward, line 28: For these types of error checking, it is easier to read the code if you do like:
```
if (!dynamic_cast<const SphericalMesh*>(mesh))
    {
    throw std::invalid_argument("Spherical geometry required");
    }

// keep going in the rest of the code
```
so that you don't have to carry an extra level of indentation. Please fix before I review the rest of this!

src/rpy_diffusive_flux.cc
```diff
@@ -0,0 +1,149 @@
...
+                        if (x==0)
+                            {
+                            flux_i(idx) = 0;
+                            }
+                        else
+                            {
+                            auto rho_x = mesh->interpolate(x, rho_i);
+                            
+                            //BD flux calculation
+                            if(i==j)
```
mphoward, line 77: This should not be an `if-else`. If `i == j`, we should do the BD flux. Then, we always do the RPY flux regardless.

src/rpy_diffusive_flux.cc
```diff
@@ -0,0 +1,149 @@
...
+        }
+    else
+        {
+        throw std::invalid_argument("Geometry is not valid for RPY hydrodynamics flux calculation");
+        }
+    }
+
+int RPYDiffusiveFlux::determineBufferShape(std::shared_ptr<State> /*state*/, const std::string& /*type*/)
+    {
+    return 1;
```
mphoward, line 128: This is still incorrect, but I'm going to stop commenting on it until you try to fix it =)

src/rpy_diffusive_flux.cc
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+double RPYDiffusiveFlux::getViscosity()
```
mphoward, line 131:
```suggestion
double RPYDiffusiveFlux::getViscosity() const
```

src/rpy_diffusive_flux.cc
```diff
@@ -0,0 +1,149 @@
...
+    {
+        return viscosity_;
+    }
+
+double RPYDiffusiveFlux::setViscosity(double viscosity)
```
mphoward, line 136:
```suggestion
void RPYDiffusiveFlux::setViscosity(double viscosity)
```

src/rpy_diffusive_flux.cc
```diff
@@ -0,0 +1,149 @@
...
+    {
+        if(viscosity==0)  
+            {
+            throw std::invalid_argument("Viscosity cannot be zero");
+            }
+        else
+            {
+            viscosity_ = viscosity;
+            return viscosity_; 
+            }
```
mphoward, line 146: Use the same pattern here:
```suggestion
        if(viscosity <= 0)  
            {
            throw std::invalid_argument("Viscosity cannot be zero");
            }
        viscosity_ = viscosity;
```

mphoward (changes requested):
Good progress, here are some comments!

src/rpy_diffusive_flux.cc
```diff
@@ -0,0 +1,166 @@
...
+                            // ERROR: can't be V=+inf and have positive rho
+                            }
+                        else if(V_j(idx) < 0)
+                            {
+                            // ERROR: can't have V=-inf ever, it is a mass sink
+                            }
+                        }
+                    no_flux = (Vidx_inf || std::isinf(V_j(idx-1)));
+                    }
+                if (!no_flux)
```
mphoward, line 68: I would do it like this so that you can remove one level of indentation (the `if-else`) below as well. Using `continue` forces the loop to go to the next mesh site.
```suggestion
                    if (Vidx_inf || std::isinf(V_j(idx-1)))
                        {
                        flux_i(idx) = 0.;
                        continue;
                        }
                    }
```

src/rpy_diffusive_flux.cc
```diff
@@ -0,0 +1,166 @@
...
+                    {
+                    const auto x = mesh->lower_bound(idx);
+                    if (x==0)
+                        {
+                        flux_i(idx) = 0;
+                        }
+                    else
```
mphoward, line 75: Same thing here, you can use continue and lose the else to avoid another level of indentation:
```suggestion
                    if (x == 0)
                        {
                        flux_i(idx) = 0.;
                        continue;
                        }
```
You should also probably add a comment line explaining why this is.

src/rpy_diffusive_flux.cc
```diff
@@ -0,0 +1,166 @@
...
+                        {
+                        auto rho_x = mesh->interpolate(x, rho_i);
+                       
+                        //BD flux calculation
+                        if(i==j)
+                            {
+                            auto dmu_ex = 0.0;
+                            if (mu_ex_j) dmu_ex += mesh->gradient(idx,mu_ex_j);
+                            if (V_j) dmu_ex += mesh->gradient(idx,V_j);
+                            double D = 1/(6*M_PI*viscosity_*a_j);
```
mphoward, line 85: You should precompute this at the loop level of type i, and then use it here so that it doesn't have to be recomputed each time. Also make it `const`.

src/rpy_diffusive_flux.cc
```diff
@@ -0,0 +1,166 @@
...
+                            flux_i(idx) += -D*(mesh->gradient(idx,rho_j) + rho_x*dmu_ex);
+                            }
+                       
+                        //RPY flux calculation   
+                        const double d_ij = a_i + a_j;
+                        
+                        //To remove the concern about the lower bound value spill over the buffer sites
+                        double lower_ig = ceil((x-d_ij-mesh->lower_bound())/mesh->step());
+                        const int ig_low = mesh->bin(std::max(lower_ig, 0.));
```
mphoward, line 94: Shouldn't this be an int and assigned this variable?
```suggestion
                        const int ig_low = ceil((x-d_ij-mesh->lower_bound())/mesh->step());
```

src/rpy_diffusive_flux.cc
```diff
@@ -0,0 +1,166 @@
...
+                        const int ig_high = floor((mesh->upper_bound()-(x+d_ij))/mesh->step());
```
mphoward, line 95: I don't think you want to do this in this way, won't it lead to rounding up? I think you want to just use the standard bin method to round down:
```suggestion
                        const int ig_high = mesh->bin(x+d_ij);
```

src/rpy_diffusive_flux.cc
```diff
@@ -0,0 +1,166 @@
...
+                        double ig = 0.;
+                        for (int ig_idx = ig_low; ig_idx <= ig_high; ++ig_idx)
```
mphoward, line 97: Didn't we decide this needs to be `<` not `<=`?

src/rpy_diffusive_flux.cc
```diff
@@ -0,0 +1,166 @@
...
+                            {
+                            const auto y = mesh->lower_bound(ig_idx);
+                            // total gradient of chemical potential
+                            double rho_dmu = mesh->gradient(ig_idx, rho_j);
+                            auto rho_y = mesh->interpolate(y, rho_j);
+                            if (mu_ex_j)
+                                rho_dmu += rho_y * mesh->gradient(ig_idx, mu_ex_j);
+                            if (V_j)
+                                rho_dmu += rho_y * mesh->gradient(ig_idx, V_j);
+                            // M will need the viscosity too
```
mphoward, line 107: This has been addressed
```suggestion
```

src/rpy_diffusive_flux.cc
```diff
@@ -0,0 +1,166 @@
...
+                            const double M = ((a_i * a_i + 3 * a_i * a_j + a_j * a_j) * 
+                                        (d_ij - x - y) * (d_ij + x - y) * (d_ij - x + y) 
+                                        * (d_ij + x + y)) / (24 * x * x * viscosity_ * d_ij * d_ij * d_ij);
+                            ig += M * rho_dmu;
+                         
+                            flux_i(idx) += -rho_x * ig * mesh->step();
+                            }
```
mphoward, line 114: This needs to go outside the integration loop:
```suggestion
                            }
                        flux_i(idx) += -rho_x * ig * mesh->step();
```
We also still need to check the sign on it.

src/rpy_diffusive_flux.cc
```diff
@@ -0,0 +1,166 @@
...
+    if(viscosity <=0)  
+        {
+        throw std::invalid_argument("Viscosity cannot be zero");
+        }
+    viscosity_ = viscosity;
+    }
+    
+int RPYDiffusiveFlux::determineBufferShape(std::shared_ptr<State> state, const std::string& /*type*/)
+    {
+    int buffer = 0;
```
mphoward, line 153: You can compute this once you do the calculation below:
```suggestion
```

src/rpy_diffusive_flux.cc
```diff
@@ -0,0 +1,166 @@
...
+    double large_R = 0;
```
mphoward, line 154: Use the diameter so you don't have to multiply by 0.5 in the loop
```suggestion
    double max_diameter = 0; 
```

src/rpy_diffusive_flux.cc
```diff
@@ -0,0 +1,166 @@
...
+    double large_R = 0; 
+    for(const auto &i : state->getTypes()) 
+        {
+        const auto a_i = 0.5*diameters_(i);
+        if(a_i>large_R)
+            {
+            large_R = a_i;
+            }
+        buffer = a_i/2+large_R/2;
+        }
```
mphoward, line 163:
```suggestion
        const auto d_i = diameters_(i);
        if(d_i > max_diameter)
            {
            max_diameter = d_i;
            }
        }
    buffer = ceil(0.5 * (diameters_(type) + max_diameter));
```
^ I'm not sure if you need ceil here, or if it would be sufficient to cast to int (i.e., floor)

src/rpy_diffusive_flux.cc
```diff
@@ -0,0 +1,166 @@
...
+void RPYDiffusiveFlux::setViscosity(double viscosity)
+    {
+    if(viscosity <=0)  
+        {
+        throw std::invalid_argument("Viscosity cannot be zero");
+        }
+    viscosity_ = viscosity;
+    }
+    
+int RPYDiffusiveFlux::determineBufferShape(std::shared_ptr<State> state, const std::string& /*type*/)
```
mphoward, line 151: You're going to need the type here:
```suggestion
int RPYDiffusiveFlux::determineBufferShape(std::shared_ptr<State> state, const std::string& type)
```

mphoward (changes requested):

src/rpy_diffusive_flux.cc
```diff
@@ -0,0 +1,166 @@
...
+                    if (V_j) dmu_ex += mesh->gradient(idx,V_j);
+                    flux_i(idx) += -D*(mesh->gradient(idx,rho_j) + rho_x*dmu_ex);
+                    }
+                
+                //RPY flux calculation   
+                const double d_ij = a_i + a_j;
+                
+                //To remove the concern about the lower bound value spill over the buffer sites
+                double lower_ig = ceil((x-d_ij-mesh->lower_bound())/mesh->step());
+                const int ig_low = mesh->bin(std::max(lower_ig, 0.));
```
mphoward, line 101: It is wrong to use bin on lower_ig, since lower_ig is an integer not a position.
```suggestion
                const int ig_low = std::max(ceil((x-d_ij-mesh->lower_bound())/mesh->step()), 0);
```

src/rpy_diffusive_flux.cc
```diff
@@ -0,0 +1,166 @@
...
+    for(const auto &i : state->getTypes()) 
+        {
+        const auto d_i = 0.5*diameters_(i);
+        if(d_i>max_diameter)
+            {
+            max_diameter = d_i;
+            }
+        }
+    buffer = ceil(0.5*(diameters_(type)+max_diameter));
+    return buffer;
```
mphoward, line 164:
```suggestion
    for(const auto &i : state->getTypes()) 
        {
        const auto d_i = diameters_(i);
        if(d_i>max_diameter)
            {
            max_diameter = d_i;
            }
        }
    return ceil(0.5*(diameters_(type)+max_diameter));
```

mphoward (changes requested):
Here are a couple small things I noticed quickly. Once you fix these, I think we should try a small test, like running a very short DDFT calculation, to see if it crashes. Then, we can go back to implement a proper unit test

include/flyft/rpy_diffusive_flux.h
```diff
@@ -0,0 +1,37 @@
...
+        void setViscosity(double viscosity);
+    
+    protected:
+        void compute(std::shared_ptr<GrandPotential> grand, std::shared_ptr<State> state) override;
+        
+        int determineBufferShape(std::shared_ptr<State> state, const std::string& type) override;
+    
+    private:
+        TypeMap<double> diameters_;
+        double viscosity_;
```
mphoward, line 33: We should add this parameter to the caching mechanism of the class. See these lines here for how this can be done:
https://github.com/mphowardlab/flyft/blob/bbd562799acac470872286128b1206a5bc7c90f0/include/flyft/wall_potential.h#L67-L87

(Note the `this->` stuff will not be needed because this isn't a templated class.)

We will also need to make some changes in the `pybind11` export. I'll make a note below.

python/flyft/_flyft/rpy_diffusive_flux.cc
```diff
@@ -0,0 +1,15 @@
...
+        .def_property_readonly("diameters",py::overload_cast<>(&RPYDiffusiveFlux::getDiameters), py::return_value_policy::reference_internal)
+        .def_property("viscosity", &RPYDiffusiveFlux::getViscosity, &RPYDiffusiveFlux::setViscosity)
```
mphoward, line 13: Do something like this here:
https://github.com/mphowardlab/flyft/blob/bbd562799acac470872286128b1206a5bc7c90f0/python/flyft/_flyft/wall_potential.h#L37-L62

python/tests/test_rpy_diffusive_flux.py
```diff
@@ -0,0 +1,164 @@
...
+    assert np.allclose(rpy.fluxes['A'], 0.)
+    
+    x = state.mesh.local.centers
+    state.fields['A'][:] = 3.0/state.mesh.full.L*(x[:])
+    grand.constrain('A', state.mesh.full.L*1.5, grand.Constraint.N)
+    rpy.compute(grand,state)
+    flux = np.zeros(len(x))
+    for i in range(0,len(x)):
+        left_edge = state.mesh.local.lower_bound(i)
+        if i==0:
```
mphoward, line 49: Why are you using these checks based on `i`, rather than the value of `left_edge`? I think you should do that so it is clearer what these conditions mean.

python/tests/test_rpy_diffusive_flux.py
```diff
@@ -0,0 +1,164 @@
...
+            M = 0
+            bd = 0
+        elif i>0 and i<10:
+            bd = 1/(3*np.pi*1.*1.)*3.0/state.mesh.full.L
+            M = left_edge*(-5+left_edge**2)/(18)*3.0/state.mesh.full.L
```
mphoward, line 54: Some comments about where these formulas / values come from would be helpful here

python/tests/test_rpy_diffusive_flux.py
```diff
@@ -0,0 +1,164 @@
...
+        else:
+            bd = 1/(3*np.pi*1.*1.)*3.0/state.mesh.full.L
+            M = -(1/18)*(5-1/(left_edge**2))*3.0/state.mesh.full.L
+        # Multiplying rho(x) into M at the end
+        flux[i] = -((3.0/state.mesh.full.L)*left_edge*M+bd)
+        print(left_edge,rpy.fluxes['A'][i],flux[i])
```
mphoward, line 60: Don't print out in tests once you think they are working

python/tests/test_rpy_diffusive_flux.py
```diff
@@ -0,0 +1,164 @@
...
+    assert np.allclose(rpy.fluxes['A'][:90], flux[:90],atol = 5e-3)
```
mphoward, line 61: Select based on edges or x, not by integer in case you change the mesh spacing later. Include a comment as to why.

python/tests/test_rpy_diffusive_flux.py
```diff
@@ -0,0 +1,164 @@
...
+    grand.constrain('A', volume, grand.Constraint.N)
+    rpy.compute(grand,state)
+    assert np.allclose(rpy.fluxes['A'], 0.)
+    
+    x = state.mesh.local.centers
+    state.fields['A'][:] = 3.0/state.mesh.full.L*(x[:])
+    grand.constrain('A', state.mesh.full.L*1.5, grand.Constraint.N)
+    rpy.compute(grand,state)
+    flux = np.zeros(len(x))
+    for i in range(0,len(x)):
```
mphoward, line 47: This is sufficient, the default is to start from 0:
```suggestion
    for i in range(len(x)):
```

python/tests/test_rpy_diffusive_flux.py
```diff
@@ -0,0 +1,164 @@
...
+        left_edge = state.mesh.local.lower_bound(i)
+        if i==0:
+            M = 0
+            bd = 0
+        elif i>0 and i<10:
+            bd = -0.031831*(1 + 3*left_edge)
```
mphoward, line 106: No magic numbers please =) Explain where the formulas come from, including a comment on how any numerical values were computed

python/tests/test_rpy_diffusive_flux.py
```diff
@@ -0,0 +1,164 @@
...
+    grand.constrain('A', volume, grand.Constraint.N)
+    grand.constrain('B', volume, grand.Constraint.N)
+    rpy.compute(grand,state)
+    flux = np.zeros(len(x))
+    for i in range(0,len(x)):
+        left_edge = state.mesh.local.lower_bound(i)
+        if i==0:
+            M = 0
+            bd = 0
+        elif i>0 and i<10:
```
mphoward, line 156: Your issues with the test cases are likely coming here, since the "special" region is different depending on which type you are looking at. I also can't follow these formulas, but shouldn't there be mobility calculations coming from both types?

mphoward (changes requested):
Thanks! This is getting close. Here are some minor comments, and then a guess as to what is going wrong in your unit test

python/tests/test_rpy_diffusive_flux.py
```diff
@@ -0,0 +1,310 @@
...
+    grand.constrain('B', 0, grand.Constraint.N)
+    rpy.compute(grand,state)
+    assert np.allclose(rpy.fluxes['A'], 0.)
+    assert np.allclose(rpy.fluxes['B'], 0.)
+
+    # same thing with all ones
+    state.fields['A'][:] = 1.
+    state.fields['B'][:] = 1.
+    grand.constrain('A', 0, grand.Constraint.N)
+    grand.constrain('B', 0, grand.Constraint.N)
```
mphoward, line 260:
```suggestion
    grand.constrain('A', state.mesh.full.volume(), grand.Constraint.N)
    grand.constrain('B', state.mesh.full.volume(), grand.Constraint.N)
```

python/tests/test_rpy_diffusive_flux.py
```diff
@@ -0,0 +1,310 @@
...
+    rpy.compute(grand,state)
+    assert np.allclose(rpy.fluxes['A'], 0.)
+    assert np.allclose(rpy.fluxes['B'], 0.)
+    
+    x = state.mesh.local.centers
+    state.fields['A'][:] = 3.0/R*(x[:])
+    state.fields['B'][:] = 3.0/R*(x[:])
+    grand.constrain('A', volume, grand.Constraint.N)
+    grand.constrain('B', volume, grand.Constraint.N)
```
mphoward, line 269: Apply same changes as above here, for these constraints, the lower bounds, etc.

src/rpy_diffusive_flux.cc
```diff
@@ -0,0 +1,175 @@
+#include "flyft/spherical_mesh.h"
+#include "flyft/rpy_diffusive_flux.h"
```
mphoward, line 2:
```suggestion
#include "flyft/rpy_diffusive_flux.h"
#include "flyft/spherical_mesh.h"
```

src/rpy_diffusive_flux.cc
```diff
@@ -0,0 +1,175 @@
...
+
+#include <exception>
+
+namespace flyft
+{
+
+RPYDiffusiveFlux::RPYDiffusiveFlux()
+    :viscosity_(1.0)
```
mphoward, line 10:
```suggestion
    : viscosity_(1.0)
```

src/rpy_diffusive_flux.cc
```diff
@@ -0,0 +1,175 @@
...
+                            {
+                            throw std::invalid_argument("Can't have V=-inf ever, it is a mass sink");
+                            }
+                        }
+                    if (Vidx_inf || std::isinf(V_j(idx-1)))
+                        {
+                        flux_i(idx) = 0.;
+                        continue;
+                        }
+                    }
```
mphoward, line 75: I think this code doesn't really make sense? Why does an infinite potential V_j mean the flux of i is zero? I think we need to think through which parts of this check are needed for all j, and which are only needed for type i. For example, I could see all of this applying when `i == j`.

src/rpy_diffusive_flux.cc
```diff
@@ -0,0 +1,175 @@
...
+                for (int ig_idx = ig_low; ig_idx < ig_high; ++ig_idx)
+                    {
+                    const auto y = mesh->lower_bound(ig_idx);
+                    // total gradient of chemical potential
+                    double rho_dmu = mesh->gradient(ig_idx, rho_j);
+                    auto rho_y = mesh->interpolate(y, rho_j);
+                    if (mu_ex_j)
+                        rho_dmu += rho_y * mesh->gradient(ig_idx, mu_ex_j);
+                    if (V_j)
+                        rho_dmu += rho_y * mesh->gradient(ig_idx, V_j);
```
mphoward, line 116: Here is where we might consider if an additional check on `V_j` being infinite is needed (this is what was causing your infs / nans before, iirc).

src/rpy_diffusive_flux.cc
```diff
@@ -0,0 +1,175 @@
...
+                    
+                    const double M = ((a_i * a_i + 3 * a_i * a_j + a_j * a_j) * 
+                                (d_ij - x - y) * (d_ij + x - y) * (d_ij - x + y) 
+                                * (d_ij + x + y)) / (24 * x * x * viscosity_ * d_ij * d_ij * d_ij);
```
mphoward, line 120: To save flops, you may want to pull the terms that don't depend on `y` outside of this loop and precompute them as a prefactor once.

src/rpy_diffusive_flux.cc
```diff
@@ -0,0 +1,175 @@
...
+const TypeMap<double>& RPYDiffusiveFlux::getDiameters() const
+    {
+    return diameters_;   
+    }
+
+double RPYDiffusiveFlux::getViscosity() const
+    {
+    return viscosity_;
+    }
+
```
mphoward, line 150:
```suggestion
```

src/rpy_diffusive_flux.cc
```diff
@@ -0,0 +1,175 @@
...
+
+void RPYDiffusiveFlux::setViscosity(double viscosity)
+    {
+    if(viscosity <=0)
```
mphoward, line 154:
```suggestion
    if(viscosity <= 0)
```

src/rpy_diffusive_flux.cc
```diff
@@ -0,0 +1,175 @@
...
+    {
+    return viscosity_;
+    }
+
+
+void RPYDiffusiveFlux::setViscosity(double viscosity)
+    {
+    if(viscosity <=0)  
+        {
+        throw std::invalid_argument("Viscosity cannot be zero");
```
mphoward, line 156:
```suggestion
        throw std::invalid_argument("Viscosity must be positive");
```

src/rpy_diffusive_flux.cc
```diff
@@ -0,0 +1,175 @@
...
+    double max_diameter = 0;
+    auto mesh = state->getMesh()->full().get();
+    for(const auto &i : state->getTypes()) 
+    {
+    const auto d_i = diameters_(i);
+    if(d_i>max_diameter)
+        {
+        max_diameter = d_i;
+        }
+    }
```
mphoward, line 172:
```suggestion
    {
        const auto d_i = diameters_(i);
        if(d_i > max_diameter)
            {
            max_diameter = d_i;
            }
    }
```

mphoward (changes requested):
Thanks, this is almost done! I think we need to have a look at the checks on the potential being infinite, but otherwise this is ready to go. I am going to commit all of the minor formatting changes, but I'd like you to have a look at the remaining comments & discuss with me so that we can finalize.

src/rpy_diffusive_flux.cc
mayukh33: I think the infs were also caused by the introduction of hardwall potential at the interface in the simulation code, even though we had the zeros boundary condition already implemented for the spherical geometry  😅

python/tests/test_rpy_diffusive_flux.py
```diff
@@ -0,0 +1,308 @@
...
+    R = state.mesh.full.L
+    
+    ig.volumes['A'] = 1.0
+    grand.ideal = ig
+    rpy.diameters['A'] = 2 * ai
+    rpy.viscosity = 1.0
+
+    # first check OK with all zeros
+    state.fields['A'][:] = 0.
+    grand.constrain('A', np.pi * R**3 * 3.0, grand.Constraint.N)
```
mphoward, line 32: This should be zero because the density is zero.
```suggestion
    grand.constrain('A', 0, grand.Constraint.N)
```

python/tests/test_rpy_diffusive_flux.py
```diff
@@ -0,0 +1,308 @@
...
+    rpy.compute(grand,state)
+    assert np.allclose(rpy.fluxes['A'], 0.)
+
+    # same thing with all ones
+    state.fields['A'][:] = 1
+    grand.constrain('A', np.pi * R**3 * 3.0, grand.Constraint.N)
```
mphoward, line 38: This should be the volume of the mesh because the density is 1
```suggestion
    grand.constrain('A', state.mesh.full.volume(), grand.Constraint.N)
```

python/tests/test_rpy_diffusive_flux.py
```diff
@@ -0,0 +1,310 @@
...
+    # same thing with all ones
+    x = state.mesh.local.centers
+    state.fields['A'][:] = 1
+    grand.constrain('A', volume, grand.Constraint.N)
+    rpy.compute(grand,state)
+    assert np.allclose(rpy.fluxes['A'], 0.)
+    
+    x = state.mesh.local.centers
+    state.fields['A'][:] = 3.0/R*(x[:])
+    grand.constrain('A', R*1.5, grand.Constraint.N)
```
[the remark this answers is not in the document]
mphoward: But this should actually have been changed as suggested! You get it by integrating the density r/3 in spherical coordinates.

python/tests/test_rpy_diffusive_flux.py
```diff
@@ -0,0 +1,308 @@
...
+    ig.volumes['A'] = 1.0
+    virial.coefficients['A','A'] = B
+    grand.ideal = ig
+    grand.excess = virial
+    rpy.diameters['A'] = 2 * ai
+    rpy.viscosity = 1
+
+    # first check OK with all zeros
+    state.fields['A'][:] = 0.
+    grand.constrain('A', np.pi * R**3 * 3.0, grand.Constraint.N)
```
mphoward, line 105: Same changes need to be made here as above on lines 105, 111, and 117.

mphoward (commented):
There are some changes that still need to be made to the tests. Remember that the `constrain` method is supposed to set the number of particles, so this is supposed to be consistent with the density field you input.

It is a bit clumsy so we should probably find a better way to do this, but for now, make sure we are consistent.

mphoward (approved):
I would like to make sure the error check on the external potential is just a simple throw, but once these changes are made, this can be merged! Thanks for the good work!

PR #41, opened 2023-09-18
Add SquareRoot parameter for spherical geometry

mayukh33:


mphoward:
This looks good, but maybe we should call this `SquareRootParameter` or something like that? To me, `QuadraticParameter` implies something like:
```
x(t) = t^2
```

mayukh33:
There is a unit test for `LinearParameter`. I will add to that.

include/flyft/parameter.h
```diff
@@ -132,7 +132,77 @@ class LinearParameter : public DoubleParameter
...
+                }
+
+            // if anything has changed, evaluate and store
+            if (this->token_.dirty())
+                {
+                if(time_ < 0)
+                    {
+                    throw std::invalid_argument("Complex numbers not allowed");
+                    }
+                value_ = sqrt(initial_*initial_+rate_*(time_-origin_));
```
mphoward, line 161: The issue isn't actually if the time is negative, it's if the argument of the square root is negative. You should do like this:
```suggestion
                const double value_sq = initial_*initial_+rate_*(time_-origin_);
                if(value_sq < 0)
                    {
                    throw std::invalid_argument("Time makes argument of square root negative");
                    }
                value_ = std::sqrt(value_sq);
```
Note I'm also using `std::sqrt`.

python/tests/test_parameter.py
```diff
@@ -23,3 +23,15 @@ def test_linear(state):
...
+    state.time = 2.0
+
+    squareroot = flyft.parameter.SquareRootParameter(3.0, 1.0, 4.5)
+    assert squareroot(state) == pytest.approx(np.sqrt(3.0*3.0+4.5*(2.0-1.0)))
+
+def test_squareroot_exception(state):
+    with pytest.raises(ValueError):
+        state.time = -1.0
+        squareroot = flyft.parameter.SquareRootParameter(3.0, 1.0, 4.5)
+        squareroot(state)
```
mphoward, line 37: You only expect a value error when you actually evaluate, so do like this:
```suggestion
    state.time = -1.0
    squareroot = flyft.parameter.SquareRootParameter(3.0, 1.0, 4.5)
    with pytest.raises(ValueError):
        squareroot(state)
```
Note also you should have a test where t is a valid (>= 0) time, but the square root cannot evaluate because the argument to the square root goes negative.

mphoward (changes requested):
Thanks! This all looks good to me. Maybe one day we can refactor this so there is less duplication between parameters, but I am good with it for now.

Can you please address the two comments below, and then we can merge?

python/tests/test_parameter.py
```diff
@@ -23,3 +23,20 @@ def test_linear(state):
...
+    squareroot = flyft.parameter.SquareRootParameter(3.0, 1.0, 4.5)
+    assert squareroot(state) == pytest.approx(np.sqrt(3.0*3.0+4.5*(2.0-1.0)))
+
+def test_squareroot_exception(state):
+    state.time = -3.0
+    squareroot = flyft.parameter.SquareRootParameter(1.0, 1.0, 1.0)
+    with pytest.raises(ValueError):
+        squareroot(state)
+    state.time = 2.0
+    squareroot = flyft.parameter.SquareRootParameter(1.0, 1.0, -3.0)
```
mphoward, line 38: One small thing here to make it simpler (& also test the setter):
```suggestion

    state.time = 2.0
    squareroot.rate = -3.0
```

mphoward (approved):
Thank you! Could you confirm the one change I marked works and commit it? This is ready to merge after you do that.

PR #48, opened 2024-01-18
Add 3D Grid interpolation for rdf and mobility integral

mayukh33:


src/three_dimension_index.cc
```diff
@@ -0,0 +1,23 @@
...
+    }
+
+int ThreeDimensionIndex::operator()(int i, int j, int k) const
+    {
+    return (i*nj_+j)*nk_+k;
+    }
+
+int ThreeDimensionIndex::size() const
+    {
+    return i*j*k;
```
mphoward, line 18:
```suggestion
    return ni_*nj_*nk_;
```

src/three_dimension_index.cc
```diff
@@ -0,0 +1,23 @@
+#include "flyft/three_dimension_index.h"
+
+
+namespace flyft
+{
+ThreeDimensionIndex::ThreeDimensionIndex(int ni , int nj, int nk):
+    ni_(ni), nj_(nj), nk_(nk)
```
mphoward, line 7:
```suggestion
ThreeDimensionIndex::ThreeDimensionIndex(int ni , int nj, int nk)
    : ni_(ni), nj_(nj), nk_(nk)
```

include/flyft/three_dimension_index.h
```diff
@@ -0,0 +1,21 @@
...
+#include<cmath>
+
+namespace flyft
+{
+class ThreeDimensionIndex
+    {
+    public:
+        ThreeDimensionIndex(int ni, int nj, int nk);
+        int size() const;
+        int operator()(int i, int j, int k) const;
```
mphoward, line 13:
```suggestion

        int operator()(int i, int j, int k) const;

        int size() const;
```

include/flyft/three_dimension_index.h
```diff
@@ -0,0 +1,21 @@
...
+    protected:
```
mphoward, line 14: These can be private, we aren't likely to subclass this
```suggestion
    private:
```

include/flyft/three_dimension_index.h
```diff
@@ -0,0 +1,21 @@
...
+        int ni_, nj_, nk_;
+        int i, j, k;
```
mphoward, line 16: These are not member variables, they are arguments of the call operator
```suggestion
```

mphoward (changes requested):

include/flyft/three_dimension_index.h
```diff
@@ -0,0 +1,27 @@
...
+        ThreeDimensionIndex(int ni, int nj, int nk);
+        int size() const;
+        
+        int operator()(int i, int j, int k) const;
+        
+        std::tuple<int, int, int> getBounds() const;
+
+        std::tuple<int, int, int> backmapping(int k) const;
+    protected:
+        int index_;
```
mphoward, line 21:
```suggestion
```

include/flyft/three_dimension_index.h
mphoward: This comment still applies

include/flyft/grid_interpolator.h
```diff
@@ -0,0 +1,32 @@
...
+
+namespace flyft
+{
+
+class GridInterpolator
+    {  
+    public:
+        GridInterpolator(std::string s);
+        GridInterpolator(std::string s, int ni, int nj, int nk);
+        GridInterpolator(std::string s,int ni, int nj, int nk, double dx, double dy, double dz);
```
mphoward, line 17: I don't think you need all of these constructors since we are just hacking this together. If the file is going to have the bounds in it as the first 6 values, then do like this:
```suggestion
        GridInterpolator(const std::string& s);
```
FYI: We use the const and & for strings to make sure it is passed by const reference.

src/grid_interpolator.cc
```diff
@@ -0,0 +1,104 @@
...
+        {
+        std::cout<<"File failed to open"<<std::endl;
+        }
+    indata>>ni>>nj>>nk>>dx>>dy>>dz;
+    indata.close();
+    GridInterpolator(s, ni, nj, nk, dx, dy, dz);
+    }    
+
+GridInterpolator::GridInterpolator(std::string s, int ni, int nj, int nk, double dx, double dy, double dz):
+    n_(ni, nj, nk), dx_(dx), dy_(dy), dz_(dz)
```
mphoward, line 30: delete all of this
```suggestion
```

src/grid_interpolator.cc
```diff
@@ -0,0 +1,104 @@
...
+                {
+                for (int j = 0; j < nj; j++)
+                    {
+                    for (int k = 0; k < nk; k++)
+                        {
+                        data_[n_(i, j, k)] = value;
+                        }
+                    }
+                }
+            }
```
mphoward, line 52: This is for sure not right, you are looping over every index in the data and assigning it value each time. So by the time this is done, the whole array will have been iterated n_.size() times, but every entry will be the last value!

It should be more like this:
```suggestion
        int idx = 0;
        while (indata >> value)
            {
            data_[idx++] = value;
            }
       if (idx != n_.size() || !indata.eof())
           {
           throw std::runtime_error("File did not have right number of elements.");
           }
```

src/grid_interpolator.cc
```diff
@@ -0,0 +1,104 @@
...
+    }    
+
+GridInterpolator::GridInterpolator(std::string s, int ni, int nj, int nk, double dx, double dy, double dz):
+    n_(ni, nj, nk), dx_(dx), dy_(dy), dz_(dz)
+    {
+    data_ = new double[n_.size()];  
+    std::ifstream indata;
+    indata.open(s);
+    while(indata.good()) 
+        {
```
mphoward, line 36: I don't think this is the right way to open & read a file, as far as I remember. I would have done:
```suggestion
    std::ifstream indata(s);
    if (!indata)
        {
        throw std::runtime_error("Failed to open file");
        }
```
Simpler initialization, no while loop needed.

src/grid_interpolator.cc
```diff
@@ -0,0 +1,104 @@
...
+        }
+    indata>>ni>>nj>>nk>>dx>>dy>>dz;
+    indata.close();
+    GridInterpolator(s, ni, nj, nk, dx, dy, dz);
+    }    
+
+GridInterpolator::GridInterpolator(std::string s, int ni, int nj, int nk, double dx, double dy, double dz):
+    n_(ni, nj, nk), dx_(dx), dy_(dy), dz_(dz)
+    {
+    data_ = new double[n_.size()];
```
mphoward, line 32: You will need to move this down, right after the file is opened, and read the 6 values in that you need like:
```
int nx, ny, nz;
indata >> nx >> ny >> nz >> dx_ >> dy_ >> dz_;
n_ = ThreeDimensionIndex(nx, ny, nz);
if (n_.size() == 0)
    {
    throw std::runtime_error("Grid has zero size");
    }
data _ = new double[n_.size()]

```

src/grid_interpolator.cc
```diff
@@ -0,0 +1,104 @@
...
+    {
+    
+    int bin_x = std::floor(x/dx_); 
+    int bin_y = std::floor(y/dy_); 
+    int bin_z = std::floor(z/dz_); 
+    int ni, nj , nk;
+    std::tie(ni, nj, nk) = n_.getBounds();
+    assert(bin_x >= 0 || bin_x < ni);
+    assert(bin_y >= 0 || bin_y < nj);
+    assert(bin_z >= 0 || bin_z < nk);
```
mphoward, line 68: both of these need to be true at the same time:
```suggestion
    assert(bin_x >= 0 && bin_x < ni);
    assert(bin_y >= 0 && bin_y < nj);
    assert(bin_z >= 0 && bin_z < nk);
```

src/grid_interpolator.cc
```diff
@@ -0,0 +1,104 @@
...
+    
+    return c0*(1-zd)+c1*zd;
+    }
+
+double* GridInterpolator::getData() const
+    {
+    return data_;
+    }
+
+GridInterpolator::~GridInterpolator()
```
mphoward, line 99: Put destructors right after constructors.

src/m.dat
mphoward, line 1: Delete this, it is research data.

src/rpy_diffusive_flux.cc
```diff
@@ -15,6 +15,9 @@ void RPYDiffusiveFlux::compute(std::shared_ptr<GrandPotential> grand, std::share
     {
     setup(grand,state);
     
+    //Interpolation of the rdf and mobility integral
+    GridInterpolator grid("m.dat");
+
```
mphoward, line 20: Let's not worry about this for now either
```suggestion
```

include/flyft/three_dimension_index.h
```diff
@@ -0,0 +1,27 @@
...
+    {
+    public:
+        ThreeDimensionIndex(int ni, int nj, int nk);
+        int size() const;
+        
+        int operator()(int i, int j, int k) const;
+        
+        std::tuple<int, int, int> getBounds() const;
+
+        std::tuple<int, int, int> backmapping(int k) const;
```
mphoward, line 19: This needs a better name for the function and its argument, but it's fine for now

src/three_dimension_index.cc
```diff
@@ -0,0 +1,34 @@
...
+std::tuple<int, int, int>ThreeDimensionIndex::getBounds() const
+    {
+    return std::make_tuple(ni_, nj_, nk_);
+    }
+
+std::tuple<int, int, int> ThreeDimensionIndex::backmapping(int k) const
+    {
+    int z = k%(nk_);
+    int y = ((k-z)/nk_)%nj_;
+    int x =  (((k-z)/nk_)-y)/nj_;
```
mphoward, line 30: Does this not work?
```suggestion
    const int z = k % nk_;
    const int y = (k / nk_) % nj_;
    const int x =  k / (n_j * nk_);
```

mphoward (changes requested):
Thanks for taking a first stab at this. The interpolator definitely needs some more work. We can chat on Monday if you can't figure it out from these comments!

src/three_dimension_index.cc
mayukh33: I tried an example hand calculation using (x,y,z) = (1, 2, 3) and (ni, nj,  nk) = (5, 5, 5), and the equations I have used work since `k-z` makes it divisible by nk while k would result in a value with decimals. I also tried with (x,y,z) = (3,1,2) with (6,6,6) and (8,4,6) grid size, and the equation I had worked for that as well.

mphoward: That is probably true, but does what I wrote also work? It uses less arithmetic and I think achieves the same result because integer math drops remainders.

include/flyft/grid_interpolator.h
mayukh33: I have tried to implement it without the constructors, but it gives me this error:
`no default constructor exists for class "flyft::ThreeDimensionIndex"C/C++(291)`, even though I give the constructor for the ThreeDimensionIndex later within the constructor.

src/three_dimension_index.cc
mayukh33: Oh, I see! Let me check!

include/flyft/grid_interpolator.h
mphoward: You need to implement a default constructor (the constructor that takes no arguments) for the ThreeDimensionIndex then! Fill the dimensions and spacings with zeros, for example

src/three_dimension_index.cc
mayukh33: Yes, it works! Sorry about that!

include/flyft/grid_interpolator.h
mayukh33: Okay!

include/flyft/grid_interpolator.h
```diff
@@ -0,0 +1,31 @@
...
+#include <tuple>
+#include <cmath>
+
+namespace flyft
+{
+
+class GridInterpolator
+    {  
+    public:
+        GridInterpolator();
```
mphoward, line 15: Unless we need it, I wouldn't give this a default constructor. It doesn't make sense to initialize this without any data with the way it currently is setup.
```suggestion
```

include/flyft/grid_interpolator.h
```diff
@@ -0,0 +1,31 @@
...
+        GridInterpolator(const std::string& s);
+        ~GridInterpolator();
+        
+        double operator()(double x, double y, double z) const;   
+        double* getData() const;
```
mphoward, line 20: You no longer need to expose this data either.
```suggestion
```

include/flyft/grid_interpolator.h
```diff
@@ -0,0 +1,31 @@
+#ifndef FLYFT_GRID_INTERPOLATOR_H_
+#define FLYFT_GRID_INTERPOLATOR_H_
+
+#include "flyft/three_dimension_index.h"
+#include <iostream>
+#include <tuple>
+#include <cmath>
```
mphoward, line 7: This is the only header you need here from the standard library, right? Include the other stuff in the `.cc` file where it's used.
```suggestion
#include <string>
```

include/flyft/rpy_diffusive_flux.h
```diff
@@ -5,6 +5,8 @@
 #include "flyft/grand_potential.h"
 #include "flyft/state.h"
 #include "flyft/type_map.h"
+#include "flyft/grid_interpolator.h"
+
```
mphoward, line 9: Don't add unrelated changes in this PR, we will deal with the RPY stuff later.
```suggestion
```

include/flyft/three_dimension_index.h
```diff
@@ -0,0 +1,28 @@
...
+        
+    private:
+        int nx_, ny_, nz_;
```
mphoward, line 23: We declare member variables one at a time
```suggestion
        int nx_;
        int ny_;
        int nz_;
```

src/grid_interpolator.cc
```diff
@@ -0,0 +1,99 @@
...
+
+#include <fstream> 
+#include <cassert>
+
+namespace flyft
+{
+GridInterpolator::GridInterpolator():
+    n_(0,0,0)
+    {
+    }
```
mphoward, line 11:
```suggestion
```

src/grid_interpolator.cc
```diff
@@ -0,0 +1,99 @@
...
+    int nx, ny, nz;
+    double dx, dy, dz;
+    
+    std::ifstream indata(s);
+    if(!indata)
+        {
+        throw std::runtime_error("File failed to open");
+        }
+        
+    indata>>nx>>ny>>nz>>dx>>dy>>dz;
```
mphoward, line 24: Use whitespace to make this more readable, and read directly into the member variables when it's possible.
```suggestion
    int nx, ny, nz;
    indata >> nx >> ny >> nz >> dx_ >> dy_ >> dz_;
```

src/grid_interpolator.cc
```diff
@@ -0,0 +1,99 @@
...
+{
+GridInterpolator::GridInterpolator():
+    n_(0,0,0)
+    {
+    }
+    
+GridInterpolator::GridInterpolator(const std::string& s)
+    {
+    int nx, ny, nz;
+    double dx, dy, dz;
```
mphoward, line 16: We put variables where we use them
```suggestion
```

src/grid_interpolator.cc
```diff
@@ -0,0 +1,99 @@
...
+    
+    std::ifstream indata(s);
+    if(!indata)
+        {
+        throw std::runtime_error("File failed to open");
+        }
+        
+    indata>>nx>>ny>>nz>>dx>>dy>>dz;
+    n_ = ThreeDimensionIndex(nx,ny,nz);
+
```
mphoward, line 26: Again, use whitespace to group concepts together. This is all related to making the indexer
```suggestion
```

src/grid_interpolator.cc
```diff
@@ -0,0 +1,99 @@
...
+    n_ = ThreeDimensionIndex(nx,ny,nz);
+    
+    if (n_.size() == 0)
+        {
+        throw std::runtime_error("Grid has zero size");
+        }
+        
+    data_ = new double[n_.size()];  
+    double value;
+
```
mphoward, line 34: Then this is all related to reading the data
```suggestion
```

src/grid_interpolator.cc
```diff
@@ -0,0 +1,99 @@
...
+        
+    data_ = new double[n_.size()];  
+    double value;
+    
+    int idx = 0;
+    while (indata >> value)
+        {
+        data_[idx++] = value;
+        }
+
```
mphoward, line 40:
```suggestion
```

src/grid_interpolator.cc
```diff
@@ -0,0 +1,99 @@
...
+        
+    if (idx != n_.size() || !indata.eof())
+        {
+        throw std::runtime_error("File did not right number of elements");
+        }
+        
+    indata.close();
+    }
+    
+
```
mphoward, line 49: One line of whitespace between functions in C++
```suggestion
```

src/grid_interpolator.cc
```diff
@@ -0,0 +1,99 @@
...
+GridInterpolator::~GridInterpolator()
+    {
+    delete[] data_;
+    }
+    
+
+double GridInterpolator::operator()(double x, double y, double z) const
+    {
+
```
mphoward, line 59: No whitespace to start a function
```suggestion
```

src/grid_interpolator.cc
```diff
@@ -0,0 +1,99 @@
...
+    {
+    
+    int bin_x = std::floor(x/dx_); 
+    int bin_y = std::floor(y/dy_); 
+    int bin_z = std::floor(z/dz_); 
+    int ni, nj , nk;
+    std::tie(ni, nj, nk) = n_.getBounds();
+    assert(bin_x >= 0 && bin_x < ni);
+    assert(bin_y >= 0 && bin_y < nj);
+    assert(bin_z >= 0 && bin_z < nk);
```
mphoward, line 67: If this is the only spot it's used, can you just take this in the tuple and then use get on it?
```suggestion
    const auto bounds = n_.getBounds();
    assert(bin_x >= 0 && bin_x < std::get<0>(bounds));
    assert(bin_y >= 0 && bin_y < std::get<1>(bounds));
    assert(bin_z >= 0 && bin_z < std::get<2>(bounds));
```

src/grid_interpolator.cc
```diff
@@ -0,0 +1,99 @@
...
+GridInterpolator::~GridInterpolator()
+    {
+    delete[] data_;
+    }
+    
+
+double GridInterpolator::operator()(double x, double y, double z) const
+    {
+    
+    int bin_x = std::floor(x/dx_);
```
mphoward, line 60: All these variables should be declared with `const` here and below, they are not modified.

src/grid_interpolator.cc
```diff
@@ -0,0 +1,99 @@
...
+    double c1 = c01*(1-yd)+c11*yd;
+    
+    return c0*(1-zd)+c1*zd;
+    }
+
+double* GridInterpolator::getData() const
+    {
+    return data_;
+    }
+
```
mphoward, line 97:
```suggestion
```

src/rpy_diffusive_flux.cc
```diff
@@ -15,6 +15,9 @@ void RPYDiffusiveFlux::compute(std::shared_ptr<GrandPotential> grand, std::share
     {
     setup(grand,state);
     
+    //Interpolation of the rdf and mobility integral
+   
+
```
mphoward, line 20: Again, make sure these files aren't even showing up in your diff
```suggestion
```

src/three_dimension_index.cc
```diff
@@ -0,0 +1,34 @@
+#include "flyft/three_dimension_index.h"
+
+
+namespace flyft
+{
```
mphoward, line 5: Where is the default constructor defined? It should be in this file, like:
```
ThreeDimensionIndex::ThreeDimensionIndex()
    : nx_(0), ny_(0), nz_(0)
    {}
```

src/three_dimension_index.cc
```diff
@@ -0,0 +1,34 @@
...
+ThreeDimensionIndex::ThreeDimensionIndex(int nx , int ny, int nz)
+    :nx_(nx), ny_(ny), nz_(nz)
+    {        
+    }
+
+int ThreeDimensionIndex::operator()(int i, int j, int k) const
```
mphoward, line 11: Remember to put functions in cc files in the same order they are declared in the header for readability

include/flyft/three_dimension_index.h
```diff
@@ -0,0 +1,28 @@
...
+    public:
+        ThreeDimensionIndex();
+        ThreeDimensionIndex(int nx, int ny, int nz);
+        int size() const;
+        
+        int operator()(int i, int j, int k) const;
+        
+        std::tuple<int, int, int> getBounds() const;
+
+        std::tuple<int, int, int> backmapping(int k) const;
```
mphoward, line 20: I suggest the following:
```suggestion
        std::tuple<int, int, int> asTuple(int idx) const;
```

src/three_dimension_index.cc
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+std::tuple<int, int, int>ThreeDimensionIndex::getBounds() const
+    {
+    return std::make_tuple(nx_, ny_, nz_);
+    }
+
+std::tuple<int, int, int> ThreeDimensionIndex::backmapping(int k) const
+    {
+    const int z = k % (nz_);
```
mphoward, line 28: Don't put parentheses around single variables
```suggestion
    const int z = k % nz_;
```

mphoward (changes requested):
Looking better, here are some more suggestions!

include/flyft/three_dimension_index.h
```diff
@@ -0,0 +1,31 @@
...
+        ThreeDimensionIndex();
+        ThreeDimensionIndex(int nx, int ny, int nz);
+        
+        int size() const;
+        
+        int operator()(int i, int j, int k) const;
+        
+        std::tuple<int, int, int> getBounds() const;
+
+        std::tuple<int, int, int> asTuple(int k) const;
```
mphoward, line 21: Rename this to `idx` everywhere. `k` is confusing because it seems like the z index.
```suggestion
        std::tuple<int, int, int> asTuple(int idx) const;
```

include/flyft/three_dimension_index.h
```diff
@@ -0,0 +1,31 @@
...
+{
+class ThreeDimensionIndex
+    {
+    public:
+        ThreeDimensionIndex();
+        ThreeDimensionIndex(int nx, int ny, int nz);
+        
+        int size() const;
+        
+        int operator()(int i, int j, int k) const;
```
mphoward, line 17: Swap the order of these, I would expect to see the call operator first since it's most important.

src/grid_interpolator.cc
mphoward: Make sure all these changes get implemented and resolve them.

src/grid_interpolator.cc
```diff
@@ -0,0 +1,82 @@
...
+    delete[] data_;
+    }
+
+double GridInterpolator::operator()(double x, double y, double z) const
+    {
+    const int bin_x = std::floor(x/dx_); 
+    const int bin_y = std::floor(y/dy_); 
+    const int bin_z = std::floor(z/dz_); 
+    
+    assert(bin_x >= 0 && bin_x < std::get<0>(n_.getBounds()));
```
mphoward, line 55: Do still store the bounds once, it is more readable. You just don't need to unpack it into 3 vars. Like so:
```suggestion
    const auto bounds = n_.getBounds();
    assert(bin_x >= 0 && bin_x < std::get<0>(bounds)));
```
apply to the next two lines as well.

src/rpy_diffusive_flux.cc
```diff
@@ -14,7 +14,6 @@ RPYDiffusiveFlux::RPYDiffusiveFlux()
 void RPYDiffusiveFlux::compute(std::shared_ptr<GrandPotential> grand, std::shared_ptr<State> state)
     {
     setup(grand,state);
```
mphoward, line 16: Please put this line break back... the easier way to make sure you actually got rid of the old changes (which is how I think you accidentally deleted this line) would have been to use git to revert the file to the main branch. You can do that now.
```
git checkout main -- rpy_diffusive_flux.cc
```

mphoward (changes requested):

mphoward (changes requested):
This looks generally good! Here are a few small things left to address.

include/flyft/three_dimensional_index.h
```diff
@@ -0,0 +1,31 @@
+#ifndef FLYFT_THREE_DIMENSIONAL_INDEX_H_
+#define FLYFT_THREE_DIMENSIONAL_INDEX_H_
+
+#include <cmath>
```
mphoward, line 4: Is this header needed here? You just have ints and tuples.

src/grid_interpolator.cc
```diff
@@ -0,0 +1,84 @@
...
+    const int bin_z = std::floor(fz); 
+    const double xd = fx - bin_x;
+    const double yd = fy - bin_y;
+    const double zd = fz - bin_z;
+    #ifndef NDEBUG
+    const auto bounds = n_.getBounds();
+    assert(bin_x >= 0 && bin_x < std::get<0>(bounds));
+    assert(bin_y >= 0 && bin_y < std::get<1>(bounds));
+    assert(bin_z >= 0 && bin_z < std::get<2>(bounds));
+    #endif
```
mphoward, line 64: whitespace for readability
```suggestion

    #ifndef NDEBUG
    const auto bounds = n_.getBounds();
    assert(bin_x >= 0 && bin_x < std::get<0>(bounds));
    assert(bin_y >= 0 && bin_y < std::get<1>(bounds));
    assert(bin_z >= 0 && bin_z < std::get<2>(bounds));
    #endif

```

CMakeLists.txt
```diff
@@ -8,7 +8,7 @@ project(
 # set default build type https://blog.kitware.com/cmake-and-the-default-build-type/
 if(NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
     message(STATUS "Setting build type to 'Release' as none was specified.")
-    set(CMAKE_BUILD_TYPE Release CACHE STRING "Choose the type of build." FORCE)
+    set(CMAKE_BUILD_TYPE RelWithDebInfo CACHE STRING "Choose the type of build." FORCE)
```
mphoward, line 11: revert

mphoward (changes requested):

mphoward (approved):
Looks great, thanks!

PR #49, opened 2024-01-30
Add interpolated rdf and mobility integral to RPY diffusive flux

mayukh33:


src/rpy_diffusive_flux.cc
```diff
@@ -33,7 +33,9 @@ void RPYDiffusiveFlux::compute(std::shared_ptr<GrandPotential> grand, std::share
         {
         throw std::invalid_argument("Spherical geometry required");
         }
-
+        
+    GridInterpolator g("/scratch2/mzk0148/projects/ddft_droplet/data/rdf_surrogate/m.dat");
```
mphoward, line 37: You will want to construct and allocate this only once. The easiest way is to add a shared pointer to the class:
```
std::shared_ptr<GridInterpolator> m_mobility;
```
and then for now, you can make it in the constructor:
```
m_mobility = std::make_shared<GridInterpolator>("/scratch2/...");
```
(We will want a better way to do that later, but this is good for now.)

src/rpy_diffusive_flux.cc
```diff
@@ -103,7 +105,10 @@ void RPYDiffusiveFlux::compute(std::shared_ptr<GrandPotential> grand, std::share
                     
                     const double M =  prefactor * (d_ij - x - y) * (d_ij + x - y) * (d_ij - x + y) 
                                         * (d_ij + x + y);
-                    ig += M * rho_dmu;
+                    const double dx = y-x;
+                    const double phi = (M_PI/6)*rho_y;                
+                    const auto M_ex = g(x,dx,phi);
+                    ig += (M + M_ex) * rho_dmu;
```
mphoward, line 111: This is not right for a few reasons:
* Doesn't the mobility in the file include the ideal part?
* The limits of integration need to be updated.
* The padding needs to also be updated accordingly.
* You need to clamp the values that are outside the bounds of your interpolant, or you will get an error.

Additionally, I recommend that for now you throw an error at the top of the function if there's more than 1 type and the diameter of that type is anything other than 1, as this will not currently work in that case. You can still leave all the for loops in place so that we don't lose that code though.

mayukh33: Yes, it contains the ideal part as well. I will change that.

include/flyft/rpy_diffusive_flux.h
```diff
@@ -21,12 +22,15 @@ class RPYDiffusiveFlux : public Flux
         
         double getViscosity() const;
         void setViscosity(double viscosity);
+        
+
```
mphoward, line 26:
```suggestion
```

src/grid_interpolator.cc
```diff
@@ -16,7 +16,7 @@ GridInterpolator::GridInterpolator(const std::string& s)
         }
         
     int nx, ny, nz;   
-    indata >> nx >> ny >> nz >> dx_ >> dy_ >> dz_;
+    indata >> x_ >> cutoff_ >> rho_ >> nx >> ny >> nz >> dx_ >> dy_ >> dz_;
```
mphoward, line 19: Yes, here I think you are thinking these are going to be the bounds of your coordinates. You don't need to read them from a file, as you can compute them from the number of points and the spacing along each dimension. If your grid contains the end points (I think it does), then you should do like this:
```suggestion
    indata >> nx >> ny >> nz >> dx_ >> dy_ >> dz_;
    bounds_ = std::make_tuple((nx-1) * dx_, (ny-1) * dy_, (nz-1) * dz_);
```

include/flyft/grid_interpolator.h
```diff
@@ -15,13 +16,17 @@ class GridInterpolator
...
         
     private:
         ThreeDimensionalIndex n_;
         double* data_;
         double dx_;
         double dy_;
         double dz_;
+        double x_;
+        double cutoff_;
+        double rho_;
```
mphoward, line 29: Why are these variables here? They seem specific to your application, not to the general interpolation problem. I think you want to declare a bounds variable instead:
```suggestion
        std::tuple<double, double, double> bounds_;
```
then compute it when you read the file.

src/grid_interpolator.cc
```diff
@@ -83,4 +83,9 @@ double GridInterpolator::operator()(double x, double y, double z) const
     
     return c0*(1-zd)+c1*zd;
     }
+
+std::tuple<double, double, double> GridInterpolator::getBounds() const
+    {
+    return std::make_tuple(x_, dx_, rho_);
```
mphoward, line 89: Then here:
```suggestion
    return bounds_;
```

src/rpy_diffusive_flux.cc
```diff
@@ -1,14 +1,20 @@
...
 namespace flyft
 {
 
 RPYDiffusiveFlux::RPYDiffusiveFlux()
     : viscosity_(1.0)
     {
+    if(!mobility_)
+        {
+        mobility_ = std::make_shared<GridInterpolator>("/scratch2/mzk0148/projects/ddft_droplet/data/rdf_surrogate/m.dat"); 
+        }
```
mphoward, line 17: If you are going to do this in the constructor, you don't need the `!mobility_` since we know it isn't initialized yet.
```suggestion
    mobility_ = std::make_shared<GridInterpolator>("/scratch2/mzk0148/projects/ddft_droplet/data/rdf_surrogate/m.dat"); 
```
You would only use that if you put this code in the `compute` method, which will be called multipled times.

src/rpy_diffusive_flux.cc
```diff
@@ -83,13 +103,14 @@ void RPYDiffusiveFlux::compute(std::shared_ptr<GrandPotential> grand, std::share
...
                 const double d_ij = a_i + a_j;
-                
+                const double x_int = std::min<double>(x, max_x);
+            
                 //To remove the concern about the lower bound value spill over the buffer sites
-                const int ig_low = std::ceil((std::abs(x-d_ij)-mesh->lower_bound())/mesh->step());
-                const int ig_high = mesh->bin(x+d_ij);
+                const double cutoff = std::get<1>(g.getBounds())*d_ij;
+                const int ig_low = (x_int >= 1)?(0):std::ceil(((d_ij-x_int)-mesh->lower_bound())/mesh->step());
+                const int ig_high = mesh->bin(x_int+cutoff);
```
mphoward, line 111: This code should not use `x_int`, but instead the real `x` to set the bounds on y. You also don't need to do this all in one line if it's confusing to read.

src/rpy_diffusive_flux.cc
```diff
@@ -83,13 +103,14 @@ void RPYDiffusiveFlux::compute(std::shared_ptr<GrandPotential> grand, std::share
  
                 //RPY flux calculation   
                 const double d_ij = a_i + a_j;
-                
+                const double x_int = std::min<double>(x, max_x);
+            
                 //To remove the concern about the lower bound value spill over the buffer sites
-                const int ig_low = std::ceil((std::abs(x-d_ij)-mesh->lower_bound())/mesh->step());
-                const int ig_high = mesh->bin(x+d_ij);
+                const double cutoff = std::get<1>(g.getBounds())*d_ij;
```
mphoward, line 109: You've already hard coded the diameter, so I don't think you should multiply by it here.
```suggestion
                const double cutoff = std::get<1>(g.getBounds());
```

src/rpy_diffusive_flux.cc
```diff
@@ -100,9 +121,13 @@ void RPYDiffusiveFlux::compute(std::shared_ptr<GrandPotential> grand, std::share
                         rho_dmu += rho_y * mesh->gradient(ig_idx, mu_ex_j);
                     if (V_j)
                         rho_dmu += rho_y * mesh->gradient(ig_idx, V_j);
+                
+                    const double rho_y_int = std::min(rho_y, max_density);
+                    const double rho_x_int = std::min(rho_x, max_density);
+                    const double dx = y-x;
+                    const double phi = (M_PI/6)*(rho_y_int+rho_x_int)/2;                        
+                    const double M = g(x_int, dx, phi);
```
mphoward, line 129: Are you using `phi` or `rho` for the interpolation? I am confused because you have written `max_density` above.

src/rpy_diffusive_flux.cc
```diff
@@ -100,9 +121,13 @@ void RPYDiffusiveFlux::compute(std::shared_ptr<GrandPotential> grand, std::share
                         rho_dmu += rho_y * mesh->gradient(ig_idx, mu_ex_j);
                     if (V_j)
                         rho_dmu += rho_y * mesh->gradient(ig_idx, V_j);
+                
+                    const double rho_y_int = std::min(rho_y, max_density);
+                    const double rho_x_int = std::min(rho_x, max_density);
```
mphoward, line 126: Don't take the min here, do it once after you take the mean.

src/rpy_diffusive_flux.cc
```diff
@@ -146,14 +171,18 @@ int RPYDiffusiveFlux::determineBufferShape(std::shared_ptr<State> state, const s
...
+        {
         const auto d_i = diameters_(i);
         if(d_i > max_diameter)
             {
             max_diameter = d_i;
             }
-    }
-    return mesh->asShape(0.5*(diameters_(type)+max_diameter));
+        }
+    return mesh->asShape(del*0.5*(diameters_(type)+max_diameter));
```
mphoward, line 186: This code is confusing because it allows for multiple types, but that is not currently supported. It will still work correctly if there's only one type and the diameter is 1.0, though, so we can leave it alone and edit later.

src/rpy_diffusive_flux.cc
```diff
@@ -33,9 +39,23 @@ void RPYDiffusiveFlux::compute(std::shared_ptr<GrandPotential> grand, std::share
...
+        }
+        
+    const GridInterpolator& g = *mobility_;
+    
+    double max_x = std::get<0>(g.getBounds());
+    double max_density = std::get<2>(g.getBounds());
+    
     for (const auto &i : state->getTypes())
         {
+        if(diameters_(i) > 1)
```
mphoward, line 55: This should be like:
```suggestion
        if(diameters_(i) != 1)
```
since anything other than 1 is not allowed (including < 1). If it triggers an error because the value is close to (but not exactly) 1, we can write this check in a slightly different way.

mphoward (changes requested):
Please have a look at these comments and let me know when you have implemented the changes!

src/rpy_diffusive_flux.cc
mayukh33: It's using `phi` for interpolation right now. However, `max_density` is written based on density. I will change it to volume fraction from density.

src/grid_interpolator.cc
mayukh33: Ohh okay! I will make the changes!

src/rpy_diffusive_flux.cc
```diff
@@ -146,14 +169,18 @@ int RPYDiffusiveFlux::determineBufferShape(std::shared_ptr<State> state, const s
...
+        {
         const auto d_i = diameters_(i);
         if(d_i > max_diameter)
             {
             max_diameter = d_i;
             }
-    }
-    return mesh->asShape(0.5*(diameters_(type)+max_diameter));
+        }
+    return mesh->asShape(cutoff*0.5*(diameters_(type)+max_diameter));
```
mphoward, line 184: With the way the code is written, this should really just be:
```suggestion
    return mesh->asShape(cutoff);
```
and you don't need any of the for loop stuff.

src/rpy_diffusive_flux.cc
```diff
@@ -83,13 +101,14 @@ void RPYDiffusiveFlux::compute(std::shared_ptr<GrandPotential> grand, std::share
...
                 //RPY flux calculation   
                 const double d_ij = a_i + a_j;
-                
+                const double x_int = std::min<double>(x, max_x);
+ 
                 //To remove the concern about the lower bound value spill over the buffer sites
-                const int ig_low = std::ceil((std::abs(x-d_ij)-mesh->lower_bound())/mesh->step());
-                const int ig_high = mesh->bin(x+d_ij);
+                const int ig_low_near_center = std::ceil(((d_ij-x)-mesh->lower_bound())/mesh->step());
+                const int ig_low = (x >= 1) ? (0): ig_low_near_center;
```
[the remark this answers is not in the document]
mayukh33: On the Python side, for generating the integral values, we have this code checking the lower limit of the integral:
```
if y < (0 if x >= 1 else 1 - x):
            continue
```
For the double integration calculation on the Python side, we did not deal with the lower limit calculations of the integral. I think we tried doing the calculation mentioned above on the Python side, but we were triggering the assert statement that `r>=1`

include/flyft/grid_interpolator.h
```diff
@@ -15,13 +16,17 @@ class GridInterpolator
         ~GridInterpolator();
         
         double operator()(double x, double y, double z) const;   
+        std::tuple<double, double, double> getLowerBounds() const ;
```
mphoward, line 19:
```suggestion
        std::tuple<double, double, double> getLowerBounds() const;
```

include/flyft/rpy_diffusive_flux.h
```diff
@@ -21,12 +22,13 @@ class RPYDiffusiveFlux : public Flux
         
         double getViscosity() const;
         void setViscosity(double viscosity);
-    
+        
     protected:
         void compute(std::shared_ptr<GrandPotential> grand, std::shared_ptr<State> state) override;
-        int determineBufferShape(std::shared_ptr<State> state, const std::string& type) override;
+        int determineBufferShape(std::shared_ptr<State> state, const std::string&) override;
```
mphoward, line 28: Revert this change. Although it is not strictly required by C++, I prefer to include the names of the variables in the header for readability.
```suggestion
        int determineBufferShape(std::shared_ptr<State> state, const std::string& type) override;
```

src/grid_interpolator.cc
```diff
@@ -14,9 +13,22 @@ GridInterpolator::GridInterpolator(const std::string& s)
...
         throw std::runtime_error("File failed to open");
         }
-        
-    int nx, ny, nz;   
-    indata >> nx >> ny >> nz >> dx_ >> dy_ >> dz_;
+    
+    double x_low, x_high, y_low, y_high, z_low, z_high;
+    int nx, ny, nz;
+    
+    indata >> x_low >> x_high >> y_low >> y_high >> z_low >> z_high >>nx >> ny >> nz;
```
mphoward, line 20:
```suggestion
    indata >> x_low >> x_high >> y_low >> y_high >> z_low >> z_high >> nx >> ny >> nz;
```

src/rpy_diffusive_flux.cc
```diff
@@ -1,14 +1,17 @@
 #include "flyft/rpy_diffusive_flux.h"
 #include "flyft/spherical_mesh.h"
 
+#include <algorithm>
```
mphoward, line 4: I can't remember everything that's in this header, but what are you using that needs it?

src/rpy_diffusive_flux.cc
```diff
@@ -33,9 +36,24 @@ void RPYDiffusiveFlux::compute(std::shared_ptr<GrandPotential> grand, std::share
...
+    
+    if(state->getNumFields() > 1)
+        {
+        throw std::invalid_argument("Only one type of particle is supported");
+        }
+        
+    const GridInterpolator& g = *mobility_;
+    
+    const double max_x = std::get<0>(g.getUpperBounds());
+    const double cutoff = std::get<1>(g.getUpperBounds());
```
mphoward, line 48: You should add a sanity check here that this lower bound of this dimension is the same as `-cutoff`, or it will lead to undefined behavior later.

src/rpy_diffusive_flux.cc
```diff
@@ -83,13 +101,14 @@ void RPYDiffusiveFlux::compute(std::shared_ptr<GrandPotential> grand, std::share
  
                 //RPY flux calculation   
                 const double d_ij = a_i + a_j;
-                
+                const double x_int = std::min<double>(x, max_x);
+ 
                 //To remove the concern about the lower bound value spill over the buffer sites
-                const int ig_low = std::ceil((std::abs(x-d_ij)-mesh->lower_bound())/mesh->step());
-                const int ig_high = mesh->bin(x+d_ij);
+                const double x_low = std::max(x - cutoff, (x >= 1) ? 0 : d_ij-x);
```
mphoward, line 107: Isn't this for y?
```suggestion
                const double y_low = std::max(x - cutoff, (x >= 1) ? 0 : d_ij-x); 
```

src/rpy_diffusive_flux.cc
```diff
@@ -142,18 +164,11 @@ void RPYDiffusiveFlux::setViscosity(double viscosity)
     viscosity_ = viscosity;
     }
     
-int RPYDiffusiveFlux::determineBufferShape(std::shared_ptr<State> state, const std::string& type)
+int RPYDiffusiveFlux::determineBufferShape(std::shared_ptr<State> state, const std::string&)
```
mphoward, line 167: I prefer to comment out the variable name like this if it is unused:
```suggestion
int RPYDiffusiveFlux::determineBufferShape(std::shared_ptr<State> state, const std::string& /* type */)
```

src/rpy_diffusive_flux.cc
```diff
@@ -100,9 +119,12 @@ void RPYDiffusiveFlux::compute(std::shared_ptr<GrandPotential> grand, std::share
                         rho_dmu += rho_y * mesh->gradient(ig_idx, mu_ex_j);
                     if (V_j)
                         rho_dmu += rho_y * mesh->gradient(ig_idx, V_j);
+
+                    const double dx = y-x;
+                    const double mean_rho_int = std::min((rho_y+rho_x)/2, max_density);
+                    const double mean_phi = mean_rho_int * (M_PI/6);
+                    const double M = g(x_int, dx, mean_phi);
```
mphoward, line 126: I think at this point it makes sense for us to update our Python script to use density rather than volume fraction, so we can remove `mean_phi`. It shouldn't be too much changes!

mphoward (changes requested):
Nice work! Here are a couple minor comments, and then one about updating the `m.dat` file to use density before we try running calculations with this.

src/grid_interpolator.cc
```diff
@@ -47,21 +59,23 @@ GridInterpolator::~GridInterpolator()
...
-    const double fz = z/dz_;
-    const int bin_x = std::floor(fx); 
-    const int bin_y = std::floor(fy); 
-    const int bin_z = std::floor(fz); 
+    const double fx = (x - std::get<0>(lower_bounds_))/dx_;
+    const double fy = (y - std::get<1>(lower_bounds_))/dy_;
+    const double fz = (z - std::get<2>(lower_bounds_))/dz_;
+    
+    const auto bounds = n_.getBounds();
+    const int bin_x = (std::abs(std::floor(fx) - (std::get<0>(bounds) - 1)) == 0) ? std::floor(fx) - 1 : std::floor(fx);
```
mphoward, line 67: You are on the right track, but I don't think this is how I would do this. Instead, I would probably do something like:
```suggestion
   int bin_x = std::floor(fx);
   if(fx == std::get<0>(bounds) && x == std::get<0>(upper_bounds_))
       {
       --bin_x;
       }
```
The second condition could be relaxed to an "is close" type tolerance if we find this is not working.

src/grid_interpolator.cc
```diff
@@ -47,21 +59,23 @@ GridInterpolator::~GridInterpolator()
...
+    const auto bounds = n_.getBounds();
+    const int bin_x = (std::abs(std::floor(fx) - (std::get<0>(bounds) - 1)) == 0) ? std::floor(fx) - 1 : std::floor(fx); 
+    const int bin_y = (std::abs(std::floor(fy) - (std::get<1>(bounds) - 1)) == 0) ? std::floor(fy) - 1 : std::floor(fy); 
+    const int bin_z = (std::abs(std::floor(fz) - (std::get<2>(bounds) - 1)) == 0) ? std::floor(fz) - 1 : std::floor(fz); 
+    
     const double xd = fx - bin_x;
     const double yd = fy - bin_y;
     const double zd = fz - bin_z;
     
     #ifndef NDEBUG
```
mphoward, line 75: You can remove this if you are now using `bounds` above for something other than debugging. The `assert` statements internally will check this `ifndef`

src/grid_interpolator.cc
```diff
@@ -47,21 +59,23 @@ GridInterpolator::~GridInterpolator()
...
-    const auto bounds = n_.getBounds();
-    assert(bin_x >= 0 && bin_x < std::get<0>(bounds));
-    assert(bin_y >= 0 && bin_y < std::get<1>(bounds));
-    assert(bin_z >= 0 && bin_z < std::get<2>(bounds));
+    assert(bin_x >= 0 && bin_x < std::get<0>(bounds) - 1);
+    assert(bin_y >= 0 && bin_y < std::get<1>(bounds) - 1);
+    assert(bin_z >= 0 && bin_z < std::get<2>(bounds) - 1);
```
mphoward, line 78: Did you try adding `assert(false);` just to make sure these are being run?

mphoward (changes requested):
Much better! Here are a few small comments.

src/grid_interpolator.cc
mayukh33: Yes! I looked at it and it was not in the correct mode. I recompiled it with the Debug mode on.

include/flyft/rpy_diffusive_flux.h
```diff
@@ -21,12 +22,13 @@ class RPYDiffusiveFlux : public Flux
         
         double getViscosity() const;
         void setViscosity(double viscosity);
-    
+
```
mphoward, line 25:
```suggestion
```

src/grid_interpolator.cc
```diff
@@ -47,22 +59,35 @@ GridInterpolator::~GridInterpolator()
...
+       --bin_x;
+       }
+    if(fy == std::get<1>(bounds) - 1 && x == std::get<1>(upper_bounds_))
+       {
+       --bin_y;
+       }
+    if(fz == std::get<2>(bounds) - 1 && x == std::get<2>(upper_bounds_))
+       {
+       --bin_z;
+       }
```
mphoward, line 82: For readability, I would interleave these like this and also add a comment on the first one about why we decrement:
```suggestion

    const auto bounds = n_.getBounds();
    // x
    int bin_x = std::floor(fx);
    if(fx == std::get<0>(bounds) - 1 && x == std::get<0>(upper_bounds_))
       {
       /* for points exactly on the upper boundary, decrease by 1 to put it in
          bin below. this makes sure we only access bins within the bounds
          during interpolation
       */
       --bin_x;
       }
    // y
    int bin_y = std::floor(fy);
    if(fy == std::get<1>(bounds) - 1 && x == std::get<1>(upper_bounds_))
       {
       --bin_y;
       }
    // z
    int bin_z = std::floor(fz);
    if(fz == std::get<2>(bounds) - 1 && x == std::get<2>(upper_bounds_))
       {
       --bin_z;
       }
```

src/rpy_diffusive_flux.cc
```diff
@@ -33,9 +35,30 @@ void RPYDiffusiveFlux::compute(std::shared_ptr<GrandPotential> grand, std::share
...
+    
+    const double max_x = std::get<0>(g.getUpperBounds());
+    const double cutoff = std::get<1>(g.getUpperBounds());
+    
+    if(std::get<1>(g.getLowerBounds()) != -cutoff)
+        {
+        throw std::invalid_argument("Cutoff bounds for integration are not valid");
+        }
+    
+    const double max_density = std::get<2>(g.getUpperBounds());
```
mphoward, line 54: For readability, I would read this at the same time as the other things, then do the error check
```suggestion
    const double max_density = std::get<2>(g.getUpperBounds());

    if(std::get<1>(g.getLowerBounds()) != -cutoff)
        {
        throw std::invalid_argument("Cutoff bounds for integration are not valid");
        }

```

mphoward (commented):
I have a couple minor cosmetic things, but then this looks like it is good to go! Have you tried running this with unit tests?

mayukh33:
> I have a couple minor cosmetic things, but then this looks like it is good to go! Have you tried running this with unit tests?

I am running the Python script to generate the flux values for the comparison. However, I made a mistake in the last run, which took around 4 hours to generate the values. I am regenerating the values and will let you know what I find! I have added the rho(x) and rho(y) terms into the integral, and the density field is `(0.8/R) * x`

src/rpy_diffusive_flux.cc
```diff
@@ -83,13 +105,14 @@ void RPYDiffusiveFlux::compute(std::shared_ptr<GrandPotential> grand, std::share
  
                 //RPY flux calculation   
                 const double d_ij = a_i + a_j;
-                
+                const double x_int = std::min<double>(x, max_x);
+ 
                 //To remove the concern about the lower bound value spill over the buffer sites
-                const int ig_low = std::ceil((std::abs(x-d_ij)-mesh->lower_bound())/mesh->step());
-                const int ig_high = mesh->bin(x+d_ij);
+                const double y_low = std::max(x - cutoff, (x >= 1) ? 0 : d_ij-x);
```
mphoward, line 111: Either make these both one, or both `d_ij`. Probably `d_ij` will be easier to generalize in future!
```suggestion
                const double y_low = std::max(x - cutoff, (x >= d_ij) ? 0 : d_ij-x); 
```

mphoward (commented):
There are two minor things open to fix, both otherwise this looks good! I am not going to approve/merge it because we still have the file path hard coded in here, but I think we are good to finalize testing on it and start to try it out.

mphoward:
> I am running the Python script to generate the flux values for the comparison. However, I made a mistake in the last run, which took around 4 hours to generate the values.

If it's taking a long time to generate the reference values, you could limit yourself to only certain x values. It isn't necessary to check every single one, you could skip every couple.

mphoward (approved):
Thank you!

PR #51, opened 2025-02-20
Enable two-dimensional array in the existing data structure

mayukh33:


src/data_layout.cc
```diff
@@ -1,25 +1,41 @@
...
+    }
+//! Overloading DataLayout constructor
+DataLayout::DataLayout(const std::vector<int>& shape) : shape_(shape) {}
 
-int DataLayout::operator()(int idx) const
+int DataLayout::operator()(const std::vector<int>& idx) const
     {
-    return idx;
+    //! Strided indexing scheme for converting
+    //! N dimensional array into one dimensional array
```
mphoward, line 20: Doxygen comments do not go in regular code, use regular comments

include/flyft/data_view.h
```diff
@@ -94,9 +98,10 @@ class DataView
         {
         }
 
-    reference operator()(int idx) const
+    reference operator()(const std::vector<int>& idx) const
         {
-        return data_[layout_(start_ + idx)];
+        std::transform(idx.begin(), idx.end(), start_, idx.begin(), std::plus<int>());
```
mphoward, line 103: What is this doing, and why can't you just pass the `idx` directly? It doesn't seem like this should do anything / work because `idx` is const.

mphoward (changes requested):
Generally a good first pass! Here are some comments that need addressing / clarification.

include/flyft/data_view.h
mayukh33: I was trying elementwise summation of the `idx` vector with `start_` to replicate the operation performed in the previous implementation of this function without using loops

include/flyft/data_layout.h
```diff
@@ -3,24 +3,39 @@
...
+#include <vector>
 
 namespace flyft
     {
+//! DataLayout
+/*!
+ * Generates data structure for the custom data types used in the software.
+ * The class provides get and set functions for the array index and implements N-dimensional array
+ * mapping into one dimensional array using strided indexing scheme.
+ */
```
mphoward, line 15: The short description should be what the object is (the class name will be part of what doxygen generates), then the long description should describe what the class does (potentially with examples longer term)
```suggestion
//! Multidimensional array layout
/*!
 * DataLayout maps an N-dimensional index to one-dimensional
 * index using row-major ordering. Negative indexes are supported
 * and interpreted as relative to the zero-index.
 */
```

include/flyft/data_layout.h
```diff
@@ -3,24 +3,39 @@
...
 class DataLayout
     {
     public:
     DataLayout();
-    explicit DataLayout(int shape_);
+    //! Constructor for the DataLayout
+    /*!
+     * \param shape_ Shape of the index array
+     */
```
mphoward, line 23: Don't include class name in short descriptions. Use periods to end parameter descriptions. Parameter name needs to match argument name.
```suggestion
    //! Constructor
    /*!
     * \param shape Number of elements in array along each dimension.
     */
```

include/flyft/data_layout.h
```diff
@@ -3,24 +3,39 @@
...
+    explicit DataLayout(const std::vector<int>& shape);
 
-    int operator()(int idx) const;
-    int shape() const;
+    // Getter for the class
+    int operator()(const std::vector<int>& idx) const;
```
mphoward, line 27: This is not a doxygen comment and not descriptive
```suggestion
    //! Map a multidimensional index to a one-dimensional index
    /*!
     * \param idx Multidimensional index.
     * \return One-dimensional index.
     *
     * Row-major ordering is used.
     */
    int operator()(const std::vector<int>& idx) const;
```

include/flyft/data_layout.h
```diff
@@ -3,24 +3,39 @@
...
+
+    // Setter for shape
+    std::vector<int> shape() const;
```
mphoward, line 30: This is not doxygen and it is also not a setter. It's OK, though, to only use a short description for getters.
```suggestion
    //! Shape of the layout in each dimension
    std::vector<int> shape() const;
```

include/flyft/data_layout.h
```diff
@@ -3,24 +3,39 @@
...
+    // Getter for shape
```
mphoward, line 31:
```suggestion

    //! Total number of elements in the layout
```

include/flyft/data_layout.h
```diff
@@ -3,24 +3,39 @@
...
     int size() const;
 
     bool operator==(const DataLayout& other) const;
     bool operator!=(const DataLayout& other) const;
```
mphoward, line 35:
```suggestion
    //! Test if two layouts are equal
    /*!
     * \return True if the layouts have the same shape.
     */
    bool operator==(const DataLayout& other) const;

    //! Test if two layouts are not equal
    bool operator!=(const DataLayout& other) const;
```

include/flyft/data_layout.h
```diff
@@ -3,24 +3,39 @@
...
 
     private:
-    int shape_;
+    std::vector<int> shape_;
```
mphoward, line 38:
```suggestion
    std::vector<int> shape_;    //!< Multidimensional shape of layout
```

include/flyft/data_view.h
mphoward, line 1: Please update this file based on the suggestions I made for `data_layout.h`

src/data_layout.cc
```diff
@@ -1,27 +1,50 @@
 #include "flyft/data_layout.h"
 
 #include <utility>
+#include <vector>
 
 namespace flyft
     {
 
-DataLayout::DataLayout() : shape_(0) {}
+// Default constructor
```
mphoward, line 9: I tend to like putting all the doxygen comments in the header. They only need to be in one spot.

src/data_layout.cc
```diff
@@ -1,27 +1,43 @@
...
 int DataLayout::size() const
     {
-    return shape();
+    int N = shape_.size();
+    int size = 1;
+    for (int i = 0; i < N; ++i)
+        {
+        size *= shape_[i];
+        }
+    return size;
```
mphoward, line 40: Could this be done with `std::accumulate` like you did in `data_view.h`? I'm good with this solution if you prefer, though!

mphoward (approved):
This looks good to go! I had one small comment: let me know if you want to make that change, and otherwise I'll merge so we can move onto the next steps.

src/data_layout.cc
mayukh33: I will change this to the same implementation as the `data_view.h` one so that it is consistent. Thank you for noticing it!

mphoward (approved):
Thanks! I just created a new branch `refactor-aniso` that we can use to stage these changes as we work on them. Please base your next branches and PRs off `refactor-aniso`, thanks!

PR #52, opened 2025-04-04
Mesh for anisotropic particles

mayukh33:
…r dimensions

### DIFF
--- a/include/flyft/mesh.h
+++ b/include/flyft/mesh.h
@@ -19,22 +19,22 @@ class Mesh
          BoundaryType lower_bc,
          BoundaryType upper_bc);
 
-    std::shared_ptr<Mesh> slice(int start, int end) const;
+    std::shared_ptr<Mesh> slice(std::vector<int> start, std::vector<int> end) const;
 
     //! Get position on the mesh, defined as center of bin
-    double center(int i) const;
+    std::vector<double> center(std::vector<int> i) const;
 
     //! Lower bound of entire mesh
-    double lower_bound() const;
+    std::vector<double> lower_bound() const;
 
     //! Get lower bound of bin
-    double lower_bound(int i) const;
+    std::vector<double> lower_bound(std::vector<int> i) const;
 
     //! Upper bound of entire mesh
-    double upper_bound() const;
+    std::vector<double> upper_bound() const;
 
     //! Get upper bound of bin
-    double upper_bound(int i) const;
+    std::vector<double> upper_bound(int i) const;
 
     //! Get surface area of lower edge of bin
     virtual double area(int i) const = 0;
@@ -46,16 +46,16 @@ class Mesh
     virtual double volume(int i) const = 0;
 
     //! Get the bin for a coordinate
-    int bin(double x) const;
+    std::vector<int> bin(double x) const;
 
     //! Length of the mesh
     double L() const;
 
     //! Shape of the mesh
-    int shape() const;
+    std::vector<int> shape() const;
 
     //! Step size of the mesh
-    double step() const;
+    std::vector<double> step() const;
 
     //! Boundary condition on lower bound of mesh
     BoundaryType lower_boundary_condition() const;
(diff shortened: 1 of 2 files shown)


include/flyft/mesh.h
```diff
@@ -19,22 +19,22 @@ class Mesh
          BoundaryType lower_bc,
          BoundaryType upper_bc);
 
-    std::shared_ptr<Mesh> slice(int start, int end) const;
+    std::shared_ptr<Mesh> slice(std::vector<int> start, std::vector<int> end) const;
```
mphoward, line 22: Make sure to pass all these vectors by const reference to avoid copies
```suggestion
    std::shared_ptr<Mesh> slice(const std::vector<int>& start, const std::vector<int>& end) const;
```

include/flyft/mesh.h
```diff
@@ -19,22 +19,22 @@ class Mesh
...
 
     //! Get position on the mesh, defined as center of bin
-    double center(int i) const;
+    std::vector<double> center(std::vector<int> i) const;
```
mphoward, line 25:
```suggestion
    std::vector<double> center(const std::vector<int>& i) const;
```

include/flyft/mesh.h
```diff
@@ -19,22 +19,22 @@ class Mesh
...
 
     //! Lower bound of entire mesh
-    double lower_bound() const;
+    std::vector<double> lower_bound() const;
 
     //! Get lower bound of bin
-    double lower_bound(int i) const;
+    std::vector<double> lower_bound(std::vector<int> i) const;
```
mphoward, line 31:
```suggestion
    std::vector<double> lower_bound(const std::vector<int>& i) const;
```

include/flyft/mesh.h
```diff
@@ -19,22 +19,22 @@ class Mesh
...
 
     //! Upper bound of entire mesh
-    double upper_bound() const;
+    std::vector<double> upper_bound() const;
 
     //! Get upper bound of bin
-    double upper_bound(int i) const;
+    std::vector<double> upper_bound(std::vector<int> i) const;
```
mphoward, line 37:
```suggestion
    std::vector<double> upper_bound(const std::vector<int>& i) const;
```

include/flyft/mesh.h
```diff
@@ -46,81 +46,86 @@ class Mesh
...
+    std::vector<double> step() const;
 
     //! Boundary condition on lower bound of mesh
     BoundaryType lower_boundary_condition() const;
 
     //! Boundary condition on upper bound of mesh
     BoundaryType upper_boundary_condition() const;
 
-    int asShape(double dx) const;
+    std::vector<int> asShape(std::vector<double> dx) const;
```
mphoward, line 66:
```suggestion
    std::vector<int> asShape(const std::vector<double>& dx) const;
```

include/flyft/mesh.h
```diff
@@ -46,81 +46,86 @@ class Mesh
...
 
-    double asLength(int shape) const;
+    std::vector<double> asLength(std::vector<int> shape) const;
```
mphoward, line 68:
```suggestion
    std::vector<double> asLength(const std::vector<int>& shape) const;
```

include/flyft/mesh.h
```diff
@@ -46,81 +46,86 @@ class Mesh
...
 
-    double integrateSurface(int idx, double j_lo, double j_hi) const;
-    double integrateSurface(int idx, const DataView<double>& j) const;
-    double integrateSurface(int idx, const DataView<const double>& j) const;
+    double integrateSurface(std::vector<int> idx, double j_lo, double j_hi) const;
+    double integrateSurface(std::vector<int> idx, const DataView<double>& j) const;
+    double integrateSurface(std::vector<int> idx, const DataView<const double>& j) const;
```
mphoward, line 72: We need to be careful here. Isn't `j` now going to need to be a vector? Or 1 vector per face of the mesh site?

include/flyft/mesh.h
```diff
@@ -46,81 +46,86 @@ class Mesh
...
 
-    double integrateVolume(int idx, double f) const;
-    double integrateVolume(int idx, const DataView<double>& f) const;
-    double integrateVolume(int idx, const DataView<const double>& f) const;
+    double integrateVolume(std::vector<int> idx, double f) const;
+    double integrateVolume(std::vector<int> idx, const DataView<double>& f) const;
+    double integrateVolume(std::vector<int> idx, const DataView<const double>& f) const;
```
mphoward, line 76:
```suggestion
    double integrateVolume(const std::vector<int>& idx, double f) const;
    double integrateVolume(const std::vector<int>& idx, const DataView<double>& f) const;
    double integrateVolume(const std::vector<int>& idx, const DataView<const double>& f) const;
```

include/flyft/mesh.h
```diff
@@ -46,81 +46,86 @@ class Mesh
...
 
     template<typename T>
     typename std::remove_const<T>::type interpolate(double x, const DataView<T>& f) const;
 
-    virtual double gradient(int idx, double f_lo, double f_hi) const = 0;
-    double gradient(int idx, const DataView<const double>& f) const;
-    double gradient(int idx, const DataView<double>& f) const;
+    virtual double gradient(std::vector<int> idx, double f_lo, double f_hi) const = 0;
+    double gradient(std::vector<int> idx, const DataView<const double>& f) const;
+    double gradient(std::vector<int> idx, const DataView<double>& f) const;
```
mphoward, line 83: Similarly here, won't the gradient now return a vector?

include/flyft/mesh.h
```diff
@@ -46,81 +46,86 @@ class Mesh
...
+    std::vector<int> start_;
 
     void validateBoundaryCondition() const;
 
     virtual std::shared_ptr<Mesh> clone() const = 0;
     };
 
 template<typename T>
-typename std::remove_const<T>::type Mesh::interpolate(double x, const DataView<T>& f) const
+typename std::remove_const<T>::type Mesh::interpolate(std::vector<double> x,
```
mphoward, line 102: Is this implementation right? Don't we need to do multidimensional interpolation?

src/mesh.cc
```diff
@@ -5,18 +5,22 @@
...
-           int shape,
+Mesh::Mesh(std::vector<double> lower_bound,
+           std::vector<double> upper_bound,
+           std::vector<int> shape,
            BoundaryType lower_bc,
            BoundaryType upper_bc)
     : lower_(lower_bound), shape_(shape), lower_bc_(lower_bc), upper_bc_(upper_bc), start_(0)
     {
-    step_ = (upper_bound - lower_bound) / shape_;
+    std::vector<double> step_;
```
mphoward, line 15: Step is already a member variable
```suggestion
```

include/flyft/mesh.h
mphoward: Based on our discussion, this interpolation method probably needs to be updated so that it is operating only on *one* coordinate, with the rest fixed to a specific bin.

include/flyft/mesh.h
```diff
@@ -8,119 +8,210 @@
...
+    std::vector<double> asLength(const std::vector<int>& shape) const;
+
+    //! Get the integral of the surface over the mesh
+    /*!
+     * \param shape Shape of the mesh
+     * \return Length of the mesh
+     */
+    double integrateSurface(const std::vector<int>& idx, double j_lo, double j_hi) const;
+    double integrateSurface(const std::vector<int>& idx, const DataView<double>& j) const;
+    double integrateSurface(const std::vector<int>& idx, const DataView<const double>& j) const;
```
mphoward, line 144: Did you think about how these are going to work in the new code? Is it going to integrate over a specific face, or all of them? And what will it integrate?

mayukh33:
Hi! I think it can be merged into `refactor-aniso`. I also wanted to make a note that I had to change the black-pre-commit-mirror.yaml's black requirement from 24.10.0 to 24.8.0. Otherwise, it gave an error that it could not find the 24.10.0 among the list of supported distributions.

